PR #56, opened 2015-05-20
add support for prefixed signature

bjrmatos:


yaronn:
Thanks @bjrmatos , this is great!

can I ask you to make one change? I think every function that gets the new prefix parameter should get it without the colon and add it inside. The reason is that in some context a "prefix" variable is used without the colon and it can be confusing if we will ever need to query it.
Also some of the functions that now get the prefix argument are extension point so some ppl may implemented them without it. So just make sure it works if prefix is not specified.

bjrmatos:
You're right... updated :)

Issue #55, opened 2015-05-20
Feature request: provide a way to place the signature somewhere besides the last child of the root node


renatoargh:
Thank you for your (great) work. This is just a suggestion (that would ease my life of course haha).

bjrmatos:
i want to work on this.. currently, there is a un-documented argument that let you specify a xmlpath to a node that will let you insert the signature before it

renatoargh:
@bjrmatos great to know, would mind providing an example?

renatoargh:
Actually I am signing a node that is deep within the XML, for example;

```xml
<root>
    <company>
        <divisions>
            I want to sign this node!
        </divisions>
        <!-- I want the signature to be here -->
    </company>
    <!-- But signature ends up here -->
</root>
```

Is it possible pass a parameter that would make the signature to be a sibling of the signed node? Thank you

bjrmatos:
sadly, currently there is no way to add the signature node after `<divisions>`, but you can add it before.

```js
var SignedXml = require('xml-crypto').SignedXml,
      fs = require('fs');

    var xml = "<root>" +
                "<company>" +
                  "<divisions>beep boop</divisions>" +
                "</company>" +
              "</root>"

    var sig = new SignedXml();
    sig.addReference("//*[local-name(.)='divisions']");
    sig.signingKey = fs.readFileSync("client.pem");
    sig.computeSignature(xml, "//*[local-name(.)='divisions']"); // will insert the signature node before <divisions>
    fs.writeFileSync("signed.xml", sig.getSignedXml());
```

output ->

````xml
<root>
    <company>
        <Signature xmlns="http://www.w3.org/2000/09/xmldsig#">....</Signature>
        <divisions>beep boop</divisions>
    </company>
    <!-- But signature ends up here -->
</root>
```

renatoargh:
This still doesn't help me as I am dealing with a very strict xml schema validation from a 3rd party service, but its good to know it anyway! Thank you for your time!!

bjrmatos:
you can add the signature at runtime anywhere you like:

```js
var DOMParser = require('xmldom').DOMParser,
      XSerializer = require('xmldom').XMLSerializer,
      SignedXml = require('xml-crypto').SignedXml,
      fs = require('fs');

var xml = "<root>" +
                "<company>" +
                  "<divisions>beep boop</divisions>" +
                "</company>" +
              "</root>";

var sig = new SignedXml();

[...]

Issue #52, opened 2015-05-11
Is documentation up to date?


bjrmatos:
Hello, i just find in the source code that a EnvelopedSignature transform is supported, but nothing of this is mentioned in the docs, also there are other things that i think are out-dated too, would you accept a pull request covering this?

yaronn:
Hi @bjrmatos . I wold love to see a PR with enhancements to the documentation!

bjrmatos:
Cool! i'll work on that

bjrmatos:
waiting for #58 to resolve

bjrmatos:
closed by #58

Issue #35, opened 2014-09-04
Signed a particular structure


marcosomma:
Hi all, 
I need to generate a XML Signature like this:

```
         <Signature xmlns="http://www.w3.org/2000/09/xmldsig#">
			<SignedInfo>
				<CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#WithComments"/>
				<SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/>
				<Reference URI="">
					<Transforms>
						<Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/>
					</Transforms>
					<DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/>
					<DigestValue>puV.............3M=</DigestValue>
				</Reference>
			</SignedInfo>
			<SignatureValue>Ghog270A................b1rJCpX3A4=</SignatureValue>
			<KeyInfo>
				<X509Data><X509Certificate>MIIEM..................4EsZ3UGyvLWK+O</X509Certificate>
				</X509Data>
				<KeyValue>
					<RSAKeyValue>
						<Modulus>MIGfMA0GCSqGSIb3DQEBAQU..........T/wIDAQAB</Modulus>
						<Exponent>AQAB</Exponent>
					</RSAKeyValue>
				</KeyValue>
			</KeyInfo>
		</Signature>
```

And I have to put it in a specific node ```<security>```.

The final result would be something like that:

```
<?xml version="1.0" encoding="UTF-8"?>
<SchufaService version="2.0" xmlns="http://www.schufa.de/siml/2.0/final">
	<Security>
		<Signature xmlns="http://www.w3.org/2000/09/xmldsig#">
			<SignedInfo>
				<CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#WithComments"/>
				<SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/>
				<Reference URI="">
					<Transforms>
						<Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/>
					</Transforms>
					<DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/>
					<DigestValue>puVw.......473M=</DigestValue>
				</Reference>
			</SignedInfo>
[...]

PR #57, opened 2015-05-30
Custom location and ability to add custom attributes to the signature

bjrmatos:
this PR adds the following features:

- `computeSignature` now accepts a options object for specify **location, prefix and custom attributes** for the signature

- Ability to add **custom attributes** to the generated signature

My use case is generate something like the following document ->

image

with this: 

```js
signer.computeSignature(xml, {
 prefix: 'ds',
 attrs: {
   Id: 'signatureKG'
   // ... more attrs here ..
 }
});
```

- Ability to **specify the location of the signature based on a reference node** (defaults to append the signature in the root node)

For example, this code signs the `<repository>` node and adds the signature **after** the `<name>` node
```js
var xml = "<root><name>xml-crypto</name><repository>github</repository></root>"
var sig = new SignedXml()

sig.signingKey = fs.readFileSync("path/to/key")
sig.addReference("//*[local-name(.)='repository']")

sig.computeSignature(xml, {
  location: {
    reference: '/root/name', // xpath expression
    action: 'after'
  }
});
```

The following actions are supported: `append, prepend, before, after`.

Let me know if you see any problems with this PR

### DIFF
--- a/lib/signed-xml.js
+++ b/lib/signed-xml.js
@@ -395,17 +395,21 @@ SignedXml.prototype.addReference = function(xpath, transforms, digestAlgorithm, 
  * Options:
  *
  * - `prefix` {String} Adds a prefix for the generated signature tags
+ * - `attrs` {Object} A hash of attributes and values `attrName: value` to add to the signature root node
  *
  */
 SignedXml.prototype.computeSignature = function(xml, opts) {
   var doc = new Dom().parseFromString(xml),
       xmlNsAttr = 'xmlns',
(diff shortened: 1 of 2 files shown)


bjrmatos:
related to #55 and #35

yaronn:
Hi @bjrmatos - great work! Can you add an explanation about the new params to the README? Something very short, no more than a few words on each.

bjrmatos:
yes, additionally i'll work on #52 :)

PR #70, opened 2015-09-04
unqualified attributes should have an empty namespace

ma314smith:
I was getting an incorrect digest value when trying to validate WS-Fed metadata.  It ended up being an issue with the ordering of unqualified attributes.

Expected Canonical Output:
```xml
<EntityDescriptor xmlns="urn:oasis:names:tc:SAML:2.0:metadata" ID="_8d1dcc18-2f1e-4a93-850b-e3a3081b3ca1" entityID="https://sts.windows.net/8bd6e98d-e212-4022-b13f-a244fab4c253/">
  <RoleDescriptor xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" protocolSupportEnumeration="http://docs.oasis-open.org/wsfed/federation/200706" xsi:type="fed:SecurityTokenServiceType">
  </RoleDescriptor>
</EntityDescriptor>  
```

Incorrect Canonical Output (note the change in order for "protocolSupportEnumeration"):
```xml
<EntityDescriptor xmlns="urn:oasis:names:tc:SAML:2.0:metadata" ID="_8d1dcc18-2f1e-4a93-850b-e3a3081b3ca1" entityID="https://sts.windows.net/8bd6e98d-e212-4022-b13f-a244fab4c253/">
  <RoleDescriptor xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance "xsi:type="fed:SecurityTokenServiceType" protocolSupportEnumeration="http://docs.oasis-open.org/wsfed/federation/200706">
  </RoleDescriptor>
</EntityDescriptor>  
```

This was happening because the default namespace was being added to unqualified attributes in exclusive-canonicalization.js:
```js
if (!attr.namespaceURI && defaultNS) { attr.namespaceURI = defaultNS; }
```

According to the spec, unqualified attributes should have an empty namespace.  See the example, here and the note about "attr2": http://www.w3.org/TR/xml-c14n#Example-SETags

This pull request removes the line that was adding the default namespace, and includes a unit test to cover this scenario.

Note that 2 of the unit tests are failing for me (same behavior on Windows and on http://www.tutorialspoint.com/codingground.htm). The failing tests are "test with a document" and "test validating SAML response".  I've verified that they were also failing without my change, so it doesn't appear to be related.  They both use the same test xml, and a quick look at the canonical output shows some random `&#xD;` values, so maybe there are some whitespace/line endings that aren't being handled properly.


bjrmatos:
seems fine to me.. all tests passed in OSX, i will test in windows to see if i get failing tests

bjrmatos:
i made some changes in order to normalize line endings across OS, could you fix your PR and re-run the tests?.. also let me know if you still see the 2 failing tests in windows, i will test on windows later in order to verify the test cases.

ma314smith:
I'm still seeing the same 2 tests failing.

bjrmatos:
all test passed for me in windows 7 and node 0.10.36 (in current xml-crypto master branch)

image

could you delete your copy of xml-crypto (and fork), download again from github and re-run the tests?

in the past i noticed that some tests fails because git convert line endings based on the configuration found in the local computer.. however with the commit that i have done yesterday this should be resolved (I just checked this on my windows PC)

bjrmatos:
also.. in my windows all test passed with your fork

ma314smith:
ok, deleting my local copy and pulling it down again resolved the issue. all tests are passing now. i think we're good.

bjrmatos:
ok, all fine for me.. @yaronn are you ok with this? according to the spec this is the correct behaviour

yaronn:
looks good to me!

bjrmatos:
done!

PR #71, opened 2015-09-13
HMAC support

fcorneli:
This patch adds HMAC-SHA1 support. Verified against Java.


bjrmatos:
thnks for contributing! all fine for me.. just need a revision from @yaronn

fcorneli:
Could this patch be included in the next release? I would like to start an implementation of DSSP:
https://www.e-contract.be/sites/dssp/dssp-specs/
for NodeJS, and really need an XML Signature library that supports HMAC-SHA1 for this.

bjrmatos:
yes! this will be included in the next release, just need a revision from the author

you could start your implementation based on your fork (which along with other patch will be the next release of xml-crypto)

fcorneli:
Thanks for the info. Will start my implementation using xml-crypto.

yaronn:
Looks good to me.
Should the Java code run in every test cycle?

fcorneli:
I used the Java project to generate an example HMAC signature. I've kept the code within the pull request as it might come in handy later on if you want to check interop with Java XML signatures, then you have an easy to start off Java project available. For the moment, I don't see the added value of running the Java tests as part of the Javascript build.

yaronn:
:thumbsup:

bjrmatos:
done! thnks @yaronn for your revision and thnks @fcorneli!

PR #74, opened 2015-10-01
npm: ignore test folder

gnawhleinad:


bjrmatos:
thnks! this should reduce the package size on npm

Issue #36, opened 2014-10-01
Support for SHA-512


martin-goulet:
The library is not supporting SHA-512. I think that it would be a good improvement. Our IdP at SunGard uses SHA-512 for SAML assertion encryption/encoding and we are not able to use passport-saml because this lack of support in sml-crypto. Happily, the changes are pretty straightforward, see the patch pasted below:
# 

--- C:/rajesh/psp/portal/node_modules/passport-saml/node_modules/xml-crypto/lib/signed-xml (2).js   Fri Sep 12 16:45:15 2014
+++ C:/rajesh/psp/portal/node_modules/passport-saml/node_modules/xml-crypto/lib/signed-xml.js   Mon Sep 22 16:19:20 2014
@@ -126,7 +126,37 @@ function RSASHA256() {

 }

+function RSASHA512() {
-  
-  /**
-  \* Sign the given string using the given key
-  *
-  */
-  this.getSignature = function(signedInfo, signingKey) {  
-    var signer = crypto.createSign("RSA-SHA512")
-    signer.update(signedInfo)  
-    var res = signer.sign(signingKey, 'base64')
-    return res
-  }
-  /**
-  \* Verify the given signature of the given string using key
-  *
-  */
-  this.verifySignature = function(str, key, signatureValue) {
-    var verifier = crypto.createVerify("RSA-SHA512")
-    verifier.update(str)
-    var res = verifier.verify(key, signatureValue, 'base64')
-    return res
-  }
  +
-  this.getAlgorithmName = function() {
-    return "http://www.w3.org/2001/04/xmldsig-more#rsa-sha512"
-  }
  +
  +}
  +
  +
  /**
  - Xml signature implementation
    *
    @@ -164,7 +194,8 @@ SignedXml.HashAlgorithms = {
  
  SignedXml.SignatureAlgorithms = {
   'http://www.w3.org/2000/09/xmldsig#rsa-sha1': RSASHA1,
-  'http://www.w3.org/2001/04/xmldsig-more#rsa-sha256': RSASHA256
-  'http://www.w3.org/2001/04/xmldsig-more#rsa-sha256': RSASHA256,
-  'http://www.w3.org/2001/04/xmldsig-more#rsa-sha512': RSASHA512
  }

PR #76, opened 2015-10-02
Support SHA512 hashing algorithm

tngan:
In #36, only signature algorithm RSASHA512 is added but not SHA512 hashing algorithm.


bjrmatos:
cool! thnks

PR #79, opened 2015-10-13
fix for reference node validation

fcorneli:
The validation logic of the reference node is incorrect. This PR fixes the issue.


bjrmatos:
good catch! thnks

Issue #66, opened 2015-08-08
Prepare for 1.0


bjrmatos:
The plan (discussed [here](https://github.com/yaronn/xml-crypto/pull/59)):
- [ ] **[deprecate]** `sig.loadSignature` (at least from the public API). 
- [ ] **[new behaviour]** `checkSignature(String: xmlDocument, [String: externalSignatureXmlDocument])` will search and validate only the first signature found (a cool feature would be let the user to validate a specific signature, vía a XPath expression).
- [ ] **[new feature]** `checkAllSignatures(String: xmlDocument, [String: externalSignatureXmlDocument])` will validate all signatures found (by default it will search inside the document for all instances of <Signature /> and validate all instances accordingly. if externalSignatureXmlDocument is specified it will search inside the external document).
- [ ] bring it back the condition: `//*[local-name(.)='SignedInfo']/*[local-name(.)='Reference' and namespace-uri(.)='http://www.w3.org/2000/09/xmldsig#']` (removed [here](https://github.com/yaronn/xml-crypto/pull/59/files#diff-e74e56dcb44165bb3bdd306d48a1fb7aL318)) to ensure a well formed xml document.. it will be ok since we have all the information that we need (namespaces) in the xml document node.
- [ ] both checkSignature and checkAllSignatures **must support the externalSignatureXmlDocument param** for external signatures. 
- [ ] **[refactor]** adapt current tests
- [x] **[fix/improve]** travis build (seems like it is broken since some time ago)
- [ ] **[add]** [debug](https://github.com/visionmedia/debug) module for logging in development
- [ ] **[remove]** hacky workaround in [this](https://github.com/yaronn/xml-crypto/commit/5f2278bd8158d23c34a31a7fdaf2ab6719b3a611) [commits](https://github.com/yaronn/xml-crypto/commit/7d69a26542e9bffeaeb2ca59ade67541f66fb853) if more information is not provided [here](https://github.com/yaronn/xml-crypto/pull/82)

:tada: The user wont need to specify the signature manually, xml-crypto will do it and will support multiple signature validation. :tada:

bjrmatos:
i will post here if i find more improvements

bjrmatos:
added a task to improve travis build

bjrmatos:
added a task to use [debug](https://github.com/visionmedia/debug) module for logging useful information in development

bjrmatos:
[...]

PR #82, opened 2015-12-11
xmlns:ds should default to http://www.w3.org/2000/09/xmldsig# if not provided

bcoe:
Testing against a major SAML implementation, I found that some implementations of the XML Signature Syntax and Processing Version 2.0 spec expect that `xmlns:ds` will have a default value of `http://www.w3.org/2000/09/xmldsig#`

xml-crypto was defaulting the missing value to `xmlns:ds=""`, which broke the hash calculation.

### DIFF
--- a/lib/exclusive-canonicalization.js
+++ b/lib/exclusive-canonicalization.js
@@ -22,8 +22,8 @@ ExclusiveCanonicalization.prototype.attrCompare = function(a,b) {
 };
 
 ExclusiveCanonicalization.prototype.nsCompare = function(a,b) {
-  var attr1 = a.prefix//+a.namespaceURI;
-  var attr2 = b.prefix//+b.namespaceURI;
+  var attr1 = a.prefix;
+  var attr2 = b.prefix;
   if (attr1 == attr2) { return 0; }
   return attr1.localeCompare(attr2);
 };
@@ -77,7 +77,7 @@ ExclusiveCanonicalization.prototype.renderNs = function(node, prefixesInScope, d
   //handle the namespaceof the node itself
   if (node.prefix) {
     if (prefixesInScope.indexOf(node.prefix)==-1) {
-      nsListToRender.push({"prefix": node.prefix, "namespaceURI": node.namespaceURI});
+      nsListToRender.push({"prefix": node.prefix, "namespaceURI": node.namespaceURI || this.defaultForPrefix(node.prefix)});
       prefixesInScope.push(node.prefix);
     }
   }
@@ -119,6 +119,14 @@ ExclusiveCanonicalization.prototype.renderNs = function(node, prefixesInScope, d
   }
 
   return {"rendered": res.join(""), "newDefaultNs": newDefaultNs};
+};
+
+ExclusiveCanonicalization.prototype.defaultForPrefix = function (prefix) {
+	var defaults = {
+		ds: 'http://www.w3.org/2000/09/xmldsig#'
+	};
+
+	return defaults[prefix];
 };
 
 ExclusiveCanonicalization.prototype.processInner = function(node, prefixesInScope, defaultNs, inclusiveNamespacesPrefixList) {
--- a/test/canonicalization-unit-tests.js
+++ b/test/canonicalization-unit-tests.js
@@ -10,7 +10,7 @@ var compare = function(test, xml, xpath, expected, inclusiveNamespacesPrefixList
     var elem = select(doc, xpath)[0]
     var can = new ExclusiveCanonicalization()
     var result = can.process(elem, { inclusiveNamespacesPrefixList: inclusiveNamespacesPrefixList }).toString()
-    
+
     test.equal(expected, result)
     test.done()
 }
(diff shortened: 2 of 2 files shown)


bjrmatos:
hmm don't know if it is correct to assume that.. maybe other implementations do that but at least i don't think that the default should live in `ExclusiveCanonicalization`, i think that the code in `ExclusiveCanonicalization` should be agnostic of XML Signatures, etc.

what do you think @yaronn?

bcoe:
Reading the spec, it seems somewhat implied that the value of xmlns:ds should have this default -- observationally, this is the assumption made by the SAML provider that I'm integrating with, and by the onelogin (another SAML IDP) Ruby and Python implementations --

Based on my interpretation of the spec, it seems more correct to default to this value than an empty string for `ds` and fits with how these two SAML IDPs calculate their signature.

Feel free to change where this default value is populated from, it would be greatly appreciated if we could land functionality along these lines though, for some integration projects we're working on at npm.

yaronn:
Thanks @bcoe for doing the research on this and submitting this PR!

I just want to make sure we're not missing something here. If I understand correctly, you are signing some xml and sending it to a third party. How does your xml have the unbound ds prefix from the first place? After all it is usually generated by the signing code. I'm also asking since &lt;ds:SignedInfo /&gt; by itself does not seem a valid xml without the explicit ds definition. Maybe the ds prefix definition appears elsewhere in the xml?

I agree that ds is a standard prefix for signatures and we will get it hard coded if that what it takes to make your (and probably others) use case to work. I just want to see if maybe some more holistic approach is required here.

bcoe:
Hey, it's actually a payload coming back from a third party SSO provider. Really wired that their SHA was calculated expecting the namespace but the payload lacked it! The folks at this corp mentioned that no one had gotten a node app to work with their SSO, I bet this tiny nit is the problem.

Thanks for your hard work on this project, can't imagine doing this from scratch.

bjrmatos:
@bcoe can you post the exact payload that you are receiving? as @yaronn mentioned seems like something is missing...

bjrmatos:
@yaronn are you ok if i move away the defaultNs for the `ds` prefix somewhere else outside `ExclusiveCanonicalization `?

yaronn:
I agree with @bjrmatos that it would be interesting to see the payload if not confidential.

Meanwhile I've merged and published xml-crypto@0.8.2 since the change is important to increase the library coverage and it doesn't have potential to cause any damage (since valid input to the method should have a local namespace).

@bjrmatos sure, you can move it outside.

bcoe:
I'll make the payload anonymous and share it.

bjrmatos:
@yaronn i have pushed the update.. a quick review before publishing would be great!

when @bcoe post the payload i will add a test for SignedXml covering that case

bcoe:
@bjrmatos @yaronn Here's the payload, I had to change the URLs as to not give away the upstream service, so the signatures will not be valid:

https://gist.github.com/bcoe/4bfc7d35be1f6cde048b

bjrmatos:
hmm the payload is a valid xml, it have the namespace for the `ds` prefix

```xml
<samlp:Response ... xmlns:ds="http://www.w3.org/2000/09/xmldsig#" .. >
``` 

i have the suspicious that the problem is in the code that verifies the signature, can you show us the code that you are using for signature validation (the code where you are using xml-crypto)?

seems that there is a problem in your code when trying to load the signature

bcoe:
@bjrmatos the problem isn't with a missing `xmlns:ds` on the `samlp:Response` itself, hold on I'm retracing my steps and will have more feedback.

bcoe:
Okay, I've recreated the problem. Note this field in the example payload I pasted:

```
<ds:SignedInfo>
```

`SignedInfo` has no `xmlns:ds`, however I believe the spec expects this to be the case. Without this patch, xml-crypto ends up creating this `canonXml`:

```xml
<ds:SignedInfo xmlns:ds="">
```

This fails the hash check when applied. The ruby library, and I believe the expectation of the upstream signer is this output:

```xml
<ds:SignedInfo xmlns:ds="http://www.w3.org/2000/09/xmldsig#">
```

This creates a payload that hashes correctly; I find this behavior to be weird myself (why wouldn't the upstream service set this attribute?) -- however, it does seem correct to add the attribute.

bjrmatos:
If you are parsing the xml correctly the `xmlns:ds` in `samlp:Response` would add that namespace to all <ds:*> tags. That is the reason for what we tought that you were using an invalid xml, but now that you shared your xml i can confirm that is valid.

xml-crypto generating `xmlns:ds=""` is the reason for what i think that there is a problem in your code.. we have saw this problem before..

Can you share your code? So i can determine the problem with more precision

bcoe:
@bjrmatos the logic that fails to propagate `xmlns` is in:

https://github.com/Clever/saml2 CC: @jefff

and

https://www.npmjs.com/package/saml20 CC: @leandrob

Perhaps it's ultimately an issue with the shared dependency `xmldom`?

Seems like a reasonable safety to populate a valid `xmlns:ds` here, rather than an empty string -- but I'd definitely be curious to know where the issue is down stream.

bcoe:
Look at the info returned from another SAML payload, I see that it's actually:

```xml
<ds:Signature xmlns:ds="http://www.w3.org/2000/09/xmldsig#">
```

That has the reference to `xmlns:ds="http://www.w3.org/2000/09/xmldsig#"`. In my example payload, it's actually:

```xml
<ds:Signature Id="uuid8e64dcfb-0151-14eb-a9ba-f803e8b9de19">
```

That is causing the the combination of `xml-crypto + saml20` and `xml-crypto + saml2` to fail.

bjrmatos:
xmldom has an issue when calling `signature.toString()`, it doesnt propagate a namespace declared in the top of the xml file..

In the latest version of xml-crypto we support a node object in `.loadSignature` to avoid this issue.

@bcoe i would appreciate if you can create a repository with all the libraries you mentioned and the code that you are using, so i can debug all the way down and test if xml-crypto is ok or not.

I still think that this patch is not necessary but i want to confirm

yaronn:
The change you did looks great @bjrmatos ! I agree the issue is most likely the way xmldom exports namespace definitions which are external to the element it currently exports / serializes.

bjrmatos:
@bcoe would you be able to set up something to debug? sorry to bother you but i really think this patch is not necessary.

bcoe:
@bjrmatos as we've identified, this solution is probably ultimately an issue with `xmldom` not propagating the namespace as we'd expect.

I greatly appreciate this patch being landed, it's allowed me to release a working SAML implementation to one of npm's largest users.

Do you thin it's the responsibility of the node-SAML libraries to propagate the missing `xmlns:ds="http://www.w3.org/2000/09/xmldsig#"`? the responsibility of `xmldom`? or an issue with `xml-crypto`?

bjrmatos:
i think is the responsability of `xml-crypto` to handle this.

the main problem is because xml-crypto let you load the signature as a string. most users rely on `xmldom` to load the signature node and use xmldom serialization to pass it to `xml-crypto` -> (`sig.loadSignature(signature.toString())`), this causes problems because the string that xml-crypto receives doesn't have the correct information about the namespace of the signature (because a bug in xmldom serialization). 

we have detected this problem before and now we accept a xmldom node object in `loadSignature`, so you can do `sig.loadSignature(signature)`, the difference is that this object has the correct namespace information of the signature so we can use it correctly.

i suggest that you (and the others SAML implementations/modules) can try passing a xmldom node object to `loadSignature` instead of a string `sig.loadSignature(signatureNodeObject)`. i'm really sure that it will resolve the problem.

FYI, the plan for 1.0 version of xml-crypto #66 is that it will handle the load of the signature node for itself, so we won't have this problem anymore and the user can forget about selecting the signature node with `xmldom`, all will be handled by `xml-crypto`

bcoe:
@bjrmatos are you proposing that:

```js
check_saml_signature = (xml, certificate, cb) ->
  doc = (new xmldom.DOMParser()).parseFromString(xml)

  signatures = xmlcrypto.xpath(doc, ".//*[local-name(.)='Signature' and namespace-uri(.)='http://www.w3.org/2000/09/xmldsig#']")
  return false unless signatures.length
  signature = signatures[0]

  sig = new xmlcrypto.SignedXml()
  sig.keyInfoProvider = getKey: -> format_pem(certificate, 'CERTIFICATE')

  sig.loadSignature signature.toString()
  return sig.checkSignature(xml)```

should be:

```js
check_saml_signature = (xml, certificate, cb) ->
  doc = (new xmldom.DOMParser()).parseFromString(xml)

  signatures = xmlcrypto.xpath(doc, ".//*[local-name(.)='Signature' and namespace-uri(.)='http://www.w3.org/2000/09/xmldsig#']")
  return false unless signatures.length
  signature = signatures[0]

  sig = new xmlcrypto.SignedXml()
  sig.keyInfoProvider = getKey: -> format_pem(certificate, 'CERTIFICATE')

  sig.loadSignature signature
  return sig.checkSignature(xml)
```

bjrmatos:
@bcoe yes! but you will need to ensure that all your dependencies (SAML modules you linked before) use this signature node in this way to ensure that nothing is getting lost somewhere

PR #91, opened 2016-02-24
Add support for REC-xml-c14n-20010315

barabansheg:
This algorithm used in new MS purchase certificates. But ExclusiveCanonicalization can check signature of it.


bjrmatos:
i'm not ok with this, technically exc14n differs from c14n, it can work on some cases but it is not compliant with the spec.

ponyesteves:
I'm implementing c14n and as said here they are not the same. The differences i' ve noticed are that c14n is aware of namespaces of al parent nodes.

Issue #95, opened 2016-03-03
Incorrect digest when xml contains xml:lang


marklagendijk:
When you have an `xml:lang` attribute in your xml the calculated digest is wrong. E.g:

``` xml
<library>
    <book>
        <name xml:lang="en">Harry Potter</name>
    </book>
</library>
```

The error works both ways, if you validate using xml-crypto itself it returns `true`.

## Reproduction
1. Go to xml-crypto/example.
2. `node example.js`
3. `xmlsec1 --verify --id-attr:Id book --pubkey-cert-pem client_public.pem result.xml`. Note that it succeeds.
4. Add `xml:lang="en"` on the `name` tag.
5. `node example.js`
6. `xmlsec1 --verify --id-attr:Id book --pubkey-cert-pem client_public.pem result.xmll`. Note that it outputs an error:

```
func=xmlSecOpenSSLEvpDigestVerify:file=digests.c:line=229:obj=sha1:subj=unknown:error=12:invalid data:data and digest do not match
FAIL
SignedInfo References (ok/all): 0/1
Manifests References (ok/all): 0/0
Error: failed to verify file "result.xml"
```

yaronn:
thanks @marklagendijk for your contribution.
I'm also not sure what the spec means. the spec here https://www.w3.org/TR/xml-exc-c14n/ says:

> implementations of this specification only render attributes from the "XML" namespace (e.g., xml:lang, xml:space, and xml:base) when they are in the subset being serialized;

based on this I believe the attribute with the xml prefix should be serialized in the example you gave.
of course sometimes most implementations behave different than the standard in which case if might be ok for us to adjust too. what do you think? any idea how common this is?

cc @bjrmatos

bjrmatos:
is there some online exc14n service? we can compare the output produced by an online service or a script in another lang (php, c#, java)

marklagendijk:
@yaronn the part of the spec you quoted talks about the attributes themselves, not about their namespace. 
The question is whether you need to explicitly define the xml namespace. As I understand it you should not do this, because this namespace is [a reserved namespace](https://www.w3.org/XML/1998/namespace) which is bound by default, and should not be explicitly defined.

I am quite sure that my fix is according to spec, or at least according to all implementations of the spec. I have tested this with several implementations:

1. `xmlsec` as mentioned in this issue.
2. `Python` my implementation replaced a Python script which signed some xml which it send to a third party. I'm not sure if this used a native Python module or some other module.
[...]

PR #96, opened 2016-03-03
Update exclusive-canonicalization.js

marklagendijk:
Fix for #95. The default namespace should not be rendered explicitly.


bjrmatos:
i'm no really sure if this change is compatible with the spec so we will wait for @yaronn to review this

Issue #43, opened 2015-03-16
Can't get XML digests to match.  Possible canonicalization problem?


archa347:
I'm having some issues using the passport-saml library, which depends on xml-crypto, in a expressjs application running on a Windows server using the IISNode package.  It is failing comparing the digest of an XML document during signature validation. It may not be xml-crypto but my own configuration, but I'm hoping you can take a look at this and tell me what might be going on. After some debugging, I've tried console logging the canonXML in validateReferences and I get this:
```
<ns2:Assertion xmlns:ns2="urn:oasis:names:tc:SAML:2.0:assertion" ID="_46a4b76336ee613e48253c243a3a0b6ba082" IssueInstant="2015-03-16T23:06:50Z" Version="2.0">&#xD;
        <ns2:Issuer Format="urn:oasis:names:tc:SAML:2.0:nameid-format:entity">REDACTED</ns2:Issuer>&#xD;
        <ns2:Subject>&#xD;
            <ns2:NameID Format="urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified">REDACTED</ns2:NameID>&#xD;
            <ns2:SubjectConfirmation Method="urn:oasis:names:tc:SAML:2.0:cm:bearer">&#xD;
                <ns2:SubjectConfirmationData NotOnOrAfter="2015-03-16T23:08:20Z" Recipient="REDACTED"></ns2:SubjectConfirmationData>&#xD;
            </ns2:SubjectConfirmation>&#xD;
        </ns2:Subject>&#xD;
        <ns2:Conditions NotBefore="2015-03-16T23:06:20Z" NotOnOrAfter="2015-03-16T23:08:20Z">&#xD;
            <ns2:AudienceRestriction>&#xD;
                <ns2:Audience>REDACTED</ns2:Audience>&#xD;
            </ns2:AudienceRestriction>&#xD;
        </ns2:Conditions>&#xD;
        <ns2:AuthnStatement AuthnInstant="2015-03-16T23:06:49Z" SessionIndex="B6VBp2KsdceV80qx2kbd6lqKDCI=kgfzlw==" SessionNotOnOrAfter="2015-03-16T23:08:20Z">&#xD;
            <ns2:AuthnContext>&#xD;
                <ns2:AuthnContextClassRef>urn:oasis:names:tc:SAML:2.0:ac:classes:Password</ns2:AuthnContextClassRef>&#xD;
            </ns2:AuthnContext>&#xD;
[...]

PR #98, opened 2016-03-04
Canonicalization of line separators

### LINKED ISSUE #65: Incorrect hash when signing an xml file with line breaks (windows)
I'll add a test case to reproduce the issue and more info later..

when the issue is resolved would be great to test in linux and osx to ensure that everything is :ok_hand:

cmordue:
Convert `\r\n` new lines to canonical form: `\n`.
Convert multiple new lines to a single one.

Fixes #65 & probably #43.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -54,9 +54,13 @@ var escaped_one_to_xml_special_map = {
 }
  
 function normalizeXmlIncludingCR(string) {
-    return string.replace(/([\&"<>\r])/g, function(str, item) {
-        return xml_special_to_escaped_one_map[item];
-    })
+  // replace \r\n lines with \n
+  // See https://www.w3.org/TR/2000/WD-xml-c14n-20000119.html#charescaping
+  // and https://www.w3.org/TR/2001/REC-xml-c14n-20010315#Examples
+  return string.replace(/(\r?\n)+/g, '\n')
+    .replace(/([\&"<>\r])/g, function(str, item) {
+      return xml_special_to_escaped_one_map[item];
+  })
 }
  
 function deNormalizeXmlIncludingCR(string) {
--- a/lib/signed-xml.js
+++ b/lib/signed-xml.js
@@ -368,7 +368,7 @@ SignedXml.prototype.loadSignature = function(signatureNode) {
   }
 
   this.signatureValue =
-    utils.findFirst(signatureNode, ".//*[local-name(.)='SignatureValue']/text()").data.replace(/\n/g, '')
+    utils.findFirst(signatureNode, ".//*[local-name(.)='SignatureValue']/text()").data.replace(/\r?\n/g, '')
 
   this.keyInfo = select(signatureNode, ".//*[local-name(.)='KeyInfo']")
 }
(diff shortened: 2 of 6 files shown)


bjrmatos:
awesome! can you update your PR to include this [change](https://github.com/auth0/xml-crypto/commit/55296d32cd523d597a403039cac7a58383485e75)?

lib/utils.js
```diff
@@ -54,9 +54,13 @@ var escaped_one_to_xml_special_map = {
 }
  
 function normalizeXmlIncludingCR(string) {
-    return string.replace(/([\&"<>\r])/g, function(str, item) {
-        return xml_special_to_escaped_one_map[item];
-    })
+  // replace \r\n lines with \n
+  // See https://www.w3.org/TR/2000/WD-xml-c14n-20000119.html#charescaping
+  // and https://www.w3.org/TR/2001/REC-xml-c14n-20010315#Examples
+  return string.replace(/(\r\n|\n)+/g, '\n')
```
bjrmatos: i think the regExp would be enough with `.replace(/\r\n/g)`, since there is no need to replace `\n` with the same `\n`

cmordue: I want to convert multiple lines into a single line so to still do that this would be slightly simpler: `/(\r?\n)+/g`

bjrmatos: :+1:

cmordue:
@bjrmatos do you have a real world example for a test for your proposal? Also, shouldn't that regex be `/(\r?\n)+/g`? since we wouldn't want to replace `\r` if it is on its own

bjrmatos:
@cmordue you're right, i think the regex should be the one that you're suggesting

i don't have any real example for that case, just trusting in the auth0's fork :) but as i say i think that the regex should be `/(\r?\n)+/g`

cmordue:
@bjrmatos I included the change you proposed and included some more integration tests.

Strangely, the change to remove `\r\n` from the signature value wasn't actually necessary (at least on node 4.2.4). The signature value would include the `\r` character and node's crypto lib would decide the signature matched with or without `\r` characters in the middle of the signature from the document. But it is more correct to remove the newlines from the signature value in the first place.

bjrmatos:
> The signature value would include the \r character and node's crypto lib would decide the signature matched with or without \r characters in the middle of the signature from the document. But it is more correct to remove the newlines from the signature value in the first place.

cool! just let me test it at home and it'll be ready for merge

bjrmatos:
@cmordue thnks for this!! you're awesome :zap:

bjrmatos:
for a new version on npm, we'll wait until #95 is resolved

yaronn:
:+1: good fix!

marklagendijk:
@cmordue @yaronn @bjrmatos I believe this fix is incorrect. The documentation referenced is not the final version of c14n 1.0. The correct documentation is:
- https://www.w3.org/TR/xml-c14n#ProcessingModel (Attribute Nodes, Text Nodes) 
- https://www.w3.org/TR/xml-c14n#Example-Chars

PR #117, opened 2016-09-08
Separate namespaces with same prefix but different URI

timotm:
To allow namespace overriding in c14n.

We have a case in SAML response verification where the server sends XML with the same namespace prefix (ds) overridden on an inner element to a different URI, extract from actual sample attached. 

[namespace_override.txt](https://github.com/yaronn/xml-crypto/files/461722/namespace_override.txt)


bjrmatos:
@timotm thnk you, this looks great. if you find more tests cases for this, please feel free to open new PR's

PR #121, opened 2016-12-06
Add possible id attribute 'id'

cava23:
I'm kind of surprised I needed this change, but I don't have control over the XML I'm receiving and need to validate its signature.


bjrmatos:
@cava23 can you add a test case? 😄

cava23:
Certainly!

bjrmatos:
published in `v0.8.5`

Issue #84, opened 2016-01-30
signed assertion + signed message + "ds" prefix defined only at the root of the XML


siacomuzzi:
Hi guys,

I have a scenario where I need to validate a saml response with the following conditions:
- Assertion was signed
- Message (saml response) was signed too
- The "ds" prefix was defined only at the root of the XML (it's not in the Signature node)

Now, time to validate both signatures with the `SignedXml` class (`sig.checkSignature`):
- For the message signature validation, everything works as expected
- But for the assertion signature validation, I have the following error:

```
[ 'invalid signature: for uri #pfxabec2b1c-3915-1115-d115-6003c7afd15e calculated digest is X but the xml to validate supplies digest Y' ]
```

Do you know if this scenario is supported for your library? I mean `signed assertion + signed message + "ds" prefix defined only at the root of the XML`

```
<?xml version="1.0" encoding="UTF-8"?>
<samlp:Response xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol" xmlns:ds="http://www.w3.org/2000/09/xmldsig#" xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Destination="http://sp.example.com/demo1/index.php?acs" ID="pfx4b172e4e-9db1-5b0b-01e9-5f50031ab711" IssueInstant="2016-01-29T15:33:31Z" Version="2.0">
   <saml:Issuer Format="urn:oasis:names:tc:SAML:2.0:nameid-format:entity">http://idp.example.com/metadata.php</saml:Issuer><ds:Signature>
  <ds:SignedInfo><ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
    <ds:SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/>
[...]

PR #147, opened 2017-12-13
Added more default namespace prefix combinations 

cmordue:
These are taken from https://www.w3.org/TR/xmldsig-core2/#sec-Versions

### DIFF
--- a/lib/signed-xml.js
+++ b/lib/signed-xml.js
@@ -247,8 +247,7 @@ SignedXml.SignatureAlgorithms = {
 SignedXml.defaultNsForPrefix = {
   ds: 'http://www.w3.org/2000/09/xmldsig#',
   dsig: 'http://www.w3.org/2000/09/xmldsig#',
-  dsig11: 'http://www.w3.org/2009/xmldsig11#',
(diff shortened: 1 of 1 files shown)


bjrmatos:
LGTM, but one question. seems like `dsig2: 'http://www.w3.org/2010/xmldsig2#'` refers to XML Signature 2.0, does this mean (technically) that it is a another algorithm/transform to support? i mean, it seems that xml nodes inside `dsig2` signature are [somehow different](https://www.w3.org/TR/xmldsig-core2/#sec-o-Simple-2.0). is there any chance that introducing the `dsig2` mapping could cause any incompatibility? (xml signature 2.0 seems to have a compatibility mode so maybe there is no compatibility issues)

cmordue:
@bjrmatos It does sound like xml signature 2.0 is backwards compatible but I'm not certain there won't be problems. I haven't tried any xml signature 2.0 docs. I'll take it out.

I've submitted this primarily because this would have made it so I didn't run into #84 for the dsig namespace (I've already resolved this issue for me by following #105 here: https://github.com/leandrob/saml20/pull/20).

cmordue:
closing since doesn't appear to be useful for others

PR #157, opened 2018-06-08
Patch for non exclusive c14n

Hinaser:
# Summary
This is a patch proposal for issues regarding to non-exclusive c14n handling bug.
This PR contains README update and corresponding unit test specs.

Issues relating:
https://github.com/yaronn/xml-crypto/issues/113
https://github.com/yaronn/xml-crypto/issues/153

# Note
It was so painful to write code to be compatible with Node.js 0.10.
Give me a praise... :tired_face:

# Optional Suggestion
As you know, there is a quick-patch in `lib/signed-xml.js` as below.  
```
  //***workaround for validating windows mobile store signatures - it uses c14n but does not state it in the transforms
  if (!hasImplicitTransforms && transforms.length==1 && transforms[0]=="http://www.w3.org/2000/09/xmldsig#enveloped-signature")
    transforms.push("http://www.w3.org/2001/10/xml-exc-c14n#")
```

I found that there are a lot of signed XML in the world with **implicit** transforms.
Not limited to "_validating windows mobile store signatures_", sometimes developers need to guess 
and apply those hidden implicit transform.

So I added option to specify such implicit transform like this:
```
new SignedXml(null, {implicitTransforms: ["http://www.w3.org/2001/10/xml-exc-c14n#"]});
```

This new option is described in README in my PR.

Then this is a suggestion.

**Can we eliminate the workaround code for windows mobile store signature?**

I know there will be an breaking impact on current xml-crypto users, but it is healthier to specify such an implicit transform into an option.

Currently, my PR preserves that workaround code. So if you are worried to have impacts on existing users, you can safely ignore my suggestion.

Best regards.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,30 @@ Note:
 The xml-crypto api requires you to supply it separately the xml signature ("&lt;Signature&gt;...&lt;/Signature&gt;", in loadSignature) and the signed xml (in checkSignature). The signed xml may or may not contain the signature in it, but you are still required to supply the signature separately.
 
 
+### Caring for Implicit transform
+If you fail to verify signed XML, then one possible cause is that there are some hidden implicit transforms(#).  
+(#) Normalizing XML document to be verified. i.e. remove extra space within a tag, sorting attributes, importing namespace declared in ancestor nodes, etc.
+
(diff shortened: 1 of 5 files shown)


LoneRifle:
Duplicate of #161, which incorporates this PR.

PR #163, opened 2018-11-02
Bugfix: a namespace in the inclusive namespace list should be treated…

gcharnock:
… where non-exclusive and thus written out on the root note of the canonical document.

According to the WC3 specification, namespaces in the inclusive namespace list should be treated as if by the exclusive algorithm. For a document subset, the exclusive algorithm copies namespaces from ancestor nodes to the root of the document in canonical from (see https://www.w3.org/TR/xml-c14n11/#Example-DocSubsets)

This patch ensures that such ancestor namespaces are treated correctly as well as adding a test for this behavior.

### DIFF
--- a/lib/signed-xml.js
+++ b/lib/signed-xml.js
@@ -450,21 +450,11 @@ SignedXml.prototype.validateReferences = function(doc) {
     }
   
     /**
-     * When canonicalization algorithm is non-exclusive, search for ancestor namespaces
-     * before validating references.
+     * Search for ancestor namespaces before validating references. Ancestor namespaces are needed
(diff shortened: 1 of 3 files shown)


lib/signed-xml.js
```diff
@@ -465,10 +465,7 @@ SignedXml.prototype.validateReferences = function(doc) {
           break;
         }
       }
-    
-      if(hasNonExcC14nTransform){
```
LoneRifle: Given that we no longer need this check anymore, you could remove lines lines 457-467 as well. The commens in 452-455 should be similarly changed

PR #171, opened 2019-02-02
Accept existing xml prefixes to avoid adding to signature

### LINKED ISSUE #170: Custom KeyInfo, avoid adding empty/unneeded namespaces
I'm trying to create a custom KeyInfo. This mostly works with something like this:

```
function AssertionKeyInfo(assertionId) {
  this.getKeyInfo = function(key, prefix) {
    return `<wsse:SecurityTokenReference wsse11:TokenType="http://docs.oasis-open.org/wss/oasis-wss-saml-token-profile-1.1#SAMLV1.1" wsu:Id="STR-${uuidv4()}" 
          xmlns:wsse11="http://docs.oasis-open.org/wss/oasis-wss-wssecurity-secext-1.1.xsd">
          <wsse:KeyIdentifier ValueType="http://docs.oasis-open.org/wss/oasis-wss-saml-token-profile-1.0#SAMLAssertionID">${assertionId}</wsse:KeyIdentifier>
      </wsse:SecurityTokenReference>`;
  };
}
```

This is the result:

```
<ds:KeyInfo>
	<wsse:SecurityTokenReference wsse11:TokenType="http://docs.oasis-open.org/wss/oasis-wss-saml-token-profile-1.1#SAMLV1.1" 
		xmlns:wsu="" wsu:Id="STR-ce7f89b3-d5c0-459c-b2e7-3fd656ff7cb9" 
		xmlns:wsse11="http://docs.oasis-open.org/wss/oasis-wss-wssecurity-secext-1.1.xsd" 
		xmlns:wsse="">
		<wsse:KeyIdentifier ValueType="http://docs.oasis-open.org/wss/oasis-wss-saml-token-profile-1.0#SAMLAssertionID">_81d5fba5c807be9e9cf60c58566349b1</wsse:KeyIdentifier>
	</wsse:SecurityTokenReference>
</ds:KeyInfo>
```

The problem is that the prefix wsu and wsse is added here.
This is not needed because the namespaces are already defined outside the signature.

Is there a way to prevent adding these namespaces?

I guess this happens because the signature is parsed as a separate document?  
`var signatureDoc = new Dom().parseFromString(this.signatureXml)`

r03:
Not sure if this is the best way, but it does work.

This adds an extra option `existingPrefixes`.
I also added this to the README and there is a test.

Fixes #170

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@
 sudo: false
 language: node_js
 node_js:
-  - "0.12"
-  - "0.10"
+  - "8.15.0"
+  - "10.15.1"
 
 before_script:
   - npm install
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@
 sudo: false
 language: node_js
 node_js:
-  - "8.15.0"
-  - "10.15.1"
+  - "0.12"
+  - "0.10"
 
 before_script:
   - npm install
--- a/README.md
+++ b/README.md
@@ -204,6 +204,7 @@ To sign xml documents:
         - `prefix` - adds this value as a prefix for the generated signature tags
         - `attrs` - a hash of attributes and values `attrName: value` to add to the signature root node
         - `location` - customize the location of the signature, pass an object with a `reference` key which should contain a XPath expression to a reference node, an `action` key which should contain one of the following values: `append`, `prepend`, `before`, `after`
+        - `existingPrefixes` - A hash of prefixes and namespaces `prefix: namespace` that shouldn't be in the signature because they already exist in the xml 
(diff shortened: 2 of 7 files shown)


r03:
Shouldn't Travis use a more recent node version? 0.10 and 0.12 are really old and not supported anymore.

r03:
If I make this work on the old node version, would you accept it? I'm happy to change it then.
I now this isn't the best solution, but I couldn't think of a better one with the current xmldom.

LoneRifle:
I would - and I was planning to make the changes in your branch if not for life getting in the way. Feel free to go ahead of me.

r03:
It should be solved now. Thanks for maintaining this library!

PR #172, opened 2019-02-08
Update signed-xml.js

### LINKED ISSUE #72: enveloped transformation + exclusive-canonicalization
Can both algorithms work together? I'm getting a [digest mismatch error](https://github.com/yaronn/xml-crypto/blob/master/lib/signed-xml.js#L315) with the following "transforms" element:

```
<ds:Transforms>
  <ds:Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/>
    <xc14n:InclusiveNamespaces xmlns:xc14n="http://www.w3.org/2001/10/xml-exc-c14n#" PrefixList="xs saml xsi"/>
  </ds:Transform>
 </ds:Transforms>
```

Thanks

lpamlie:
fix namespace issue, fixes #72

### DIFF
--- a/lib/signed-xml.js
+++ b/lib/signed-xml.js
@@ -478,6 +478,30 @@ SignedXml.prototype.validateReferences = function(doc) {
 
     var hash = this.findHashAlgorithm(ref.digestAlgorithm)
     var digest = hash.getHash(canonXml)
+    
+    if (!validateDigestValue(digest, ref.digestValue)) {
+      if (ref.inclusiveNamespacesPrefixList) {
+        // fallback: apply InclusiveNamespaces workaround (https://github.com/yaronn/xml-crypto/issues/72)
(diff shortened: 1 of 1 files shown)


LoneRifle:
I guess this is a commit that was cherry-picked from auth0/xml-crypto@25a3d1c . Thanks for bringing this up; I'll look into this and other commits that their fork has and try to reconcile the codebases

lpamlie:
Yes, that is correct.  We're using samlify and got the same problem.  This commit fixes the problem.

Thanks
LP


-------- Original message --------

LoneRifle:
If it's okay with you, I will like to make a PR that merges the original commit from Auth0, which will mean that yours will be closed. I'm doing that largely to ensure that contributions from forks of this repo are accounted for when merged back here. I will give you credit for bringing this up, so hope that's okay with you.

lpamlie:
Sure, no problem.

Thanks
LP

PR #174, opened 2019-02-11
fix for enveloped signatures

fcorneli:
This patch really makes sure that we only cut out the signature in which our enveloped signature transformation is defined. For the moment this is based on the signature value...


fcorneli:
I guess this will also fix a lot of the SAML signature issues being reported.

LoneRifle:
Thanks - let me publish this out now..

Issue #167, opened 2019-01-15
Invalid signature after passport-saml upgraded xml-crypto to ^1.0.2: but solved


huineng:
I was using passport-saml since a long time and all was working fine until they upgraded crypto-xml to ^1.0.2. (from xml-crypto@0.10.1)

After 2 days banging my head on this i solved it, though i like to know if the changes from xml-crypto@0.10.1

it's all about the rsaml response attributes (examples below taken from various copies)

attributes on the incoming saml
```
<samlp:Response 
    Destination="blablabla"
    ID="FIMRSP_50b75184-0168-1d14-b75b-951f7c61a75a" 
    InResponseTo="_78262119f68b5424897b"
    IssueInstant="2019-01-15T08:53:37Z" Version="2.0" 
    xmlns:ds="http://www.w3.org/2000/09/xmldsig#"
    xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion"
    xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol" 
    xmlns:xs="http://www.w3.org/2001/XMLSchema"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
```

using crypto-xml 0.10.1 in the function to get the canonXML this produced
```
<saml:Assertion
	xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion"
	xmlns:xs="http://www.w3.org/2001/XMLSchema"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
        ID="Assertion-uuid505be4ab-0168-12f2-b56e-e5decbef3937"
        IssueInstant="2019-01-15T07:13:46Z" 
        Version="2.0">
```

with ^1.0.2 i suddenly got invalid signature, but the reason was that the digest no longer matched (invalid signature: for uri....)

the Assertion now produced
```
<saml:Assertion
	xmlns:ds="http://www.w3.org/2000/09/xmldsig#"
	xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion"
	xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol"
	xmlns:xs="http://www.w3.org/2001/XMLSchema"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
        ID="Assertion-uuid50634b8a-0168-1c5e-b8fc-e5decbef3937" 
        IssueInstant="2019-01-15T07:21:51Z" 
[...]

PR #179, opened 2019-03-18
Xml enc c14# inclusivenamespace fixes

### LINKED ISSUE #72: enveloped transformation + exclusive-canonicalization

### LINKED ISSUE #92: Digest calculation with Java endpoint
I am connecting to a java endpoint that calculates the digest in a different way. I believe it has something to do with an option to sign the Timestamp element itself or the Timestamp content.

For the following xml fragment:

```
<wsu:Timestamp wsu:Id="TS-f3c103e9-1897-43d8-8cf6-274bdb647678">
  <wsu:Created>2016-02-24T15:32:12.693Z</wsu:Created>
  <wsu:Expires>2016-02-24T15:37:12.693Z</wsu:Expires>
</wsu:Timestamp>
```

Java produces the following transform:

```
<ds:Reference URI="#TS-f3c103e9-1897-43d8-8cf6-274bdb647678">
    <ds:Transforms>
          <ec:InclusiveNamespaces xmlns:ec="http://www.w3.org/2001/10/xml-exc-c14n#" PrefixList="wsse soap"/>
       </ds:Transform>
   </ds:Transforms>
   <ds:DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/>
   <ds:DigestValue>1rjXQQWVMM5KBWY8uswUynk6PCk=</ds:DigestValue>
</ds:Reference>
```

However, the same xml fragment produces a different digest value using xml-crypto. I am assuming it has something to do with canonicalization between node and java.

In java, there is an option to sign the "contents" or the "element". By default it signs the contents, although its difficult to tell what that really means.

Is there a known issue when interop between java/soap and other libs? Is there someway to solve the mismatch?

### LINKED ISSUE #94: InclusiveNamespaces are not considered in the Canonicalization algorithm
Hi I have read the other issues but haven't found a solution to this problem.

I am trying to validate a signature that has `InclusiveNamespaces` as a child of the `CanonicalizationMethod` like this:

```
<ds:CanonicalizationMethod` Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#">
  <ec:InclusiveNamespaces xmlns:ec="http://www.w3.org/2001/10/xml-exc-c14n#" PrefixList="soap"/>
</ds:CanonicalizationMethod>
```

This should generate a canonicalized form with the namespace included, like this:

```
<ds:SignedInfo xmlns:ds="http://www.w3.org/2000/09/xmldsig#" xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/">
  <ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#">
    <ec:InclusiveNamespaces xmlns:ec="http://www.w3.org/2001/10/xml-exc-c14n#" PrefixList="soap" />
  </ds:CanonicalizationMethod>
  <ds:SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1" />
...
...
```

but is only generating the `ds:SignedInfo` node with only one namespace (I think in the code is referenced as default `defaultForPrefix`, like this:

```
<ds:SignedInfo xmlns:ds="http://www.w3.org/2000/09/xmldsig#">
  <ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#">
    <ec:InclusiveNamespaces xmlns:ec="http://www.w3.org/2001/10/xml-exc-c14n#" PrefixList="soap" />
  </ds:CanonicalizationMethod>
  <ds:SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1" />
...
...
```

Is there a workaround for this? Thanks in advance

### LINKED ISSUE #145: signature verification fails on 1 out of 3 messages
hi,

i have a strange issue with verifying signature with a specific xml message.
i've created a repro code on runkit available here: https://runkit.com/phra/59f0d6670f3020001512de20

as you can see the first two messages get validated correctly while the last one does not. do you have idea of what can be? (i've used the `Array.prototype.pop()` method on the result of the xpath query in order to retrieve the last signature and not the inner ones)

thanks.

bazzadp:
This is a more generic fix than https://github.com/yaronn/xml-crypto/pull/172 which had several issues in my opinion:

1. It hardcoded in SAML name spaces so didn't work with other XML types with similar issue
2. It returned true after the first valid digest was encountered so messages with multiple digests were not properly checked.
3. It seemed to check the digest 3 times which is a bit overkill.

This pull request refactors the code to make it more generic meaning it should fix the original issue (https://github.com/yaronn/xml-crypto/issues/72) but also https://github.com/yaronn/xml-crypto/issues/94, https://github.com/yaronn/xml-crypto/issues/145 and also https://github.com/yaronn/xml-crypto/issues/92. It would be good if all of those who raised this could confirm if this does fix their issues.

I also added a test case for this for a SOAP message from SOAP-UI which likes to use these extra namespaces (my original problem and why I started looking at this!).

There's a bit of refactoring in this and I didn't see why ancestors were only being calculated for non-exclusive canonicalization when this problem shows they are also needed for exclusive. Perhaps I have missed something here? All the test cases still pass (including the new one) with this fix but not sure how comprehensive they are so would be good to get others input on this.

Let me know your thoughts,
Barry

***
Fixes #92
Fixes #94
Fixes #72
Fixes #145


LoneRifle:
Hey @bazzadp , thanks for the PR. Apologies for not being able to review this in good time due to other commitments. 

To answer your question - #163 was originally filed to address ancestors used in canonicalization, but caused issue #167 and was hence backed out. I'm hoping that what you have done has properly fixed this, and will review in good time!

If you wish, I can also give you publishing rights to this package so that there is at least one other person who can do so.

bazzadp:
Hi @LoneRifle thanks for getting back.

Will have a more detailed look at https://github.com/yaronn/xml-crypto/pull/163 but from a (very brief!) look it seems to do a lot of the same thing, but does it in exclusive-canonicalization.js rather than where I've done it in signed-xml.js. To be honest it probably does more belong in with the canonicalization code, than signed-xml code, so maybe I should refactor the code before you accept it? Or if this works, do we go with it for now, and refactor later?

Do we understand why that previous commit failed and does a test case exist to confirm if my code has the same issue?

If you're looking for help to maintain this package then I can possibly give some time, but definitely not approving my own pull request as need a second pair of eyes on it :-)

LoneRifle:
I've not been able to look further into the problem, and unfortunately a test case that identifies the problem does not exist. Happy for you to refactor as you deem fit. I'll certainly look into the PR as a matter of responsibility, we shouldn't be approving our own PRs =)

bazzadp:
@LoneRifle ok the good news is the https://github.com/yaronn/xml-crypto/pull/163 breaks the test suite for two tests:

1. My new test (valid_signature_with_unused_prefixes.xml)
2. "signer adds existing prefixes" - this was added with https://github.com/yaronn/xml-crypto/pull/171 so didn't exist when https://github.com/yaronn/xml-crypto/pull/163 was originally submitted.

My change passes the test suite, so that gives me greater confidence that my test hopefully won't have the same issue. However, without a test case for that it's difficult to be sure though. Have asked for help on https://github.com/yaronn/xml-crypto/issues/167 to reduce likelihood of issues.

I also looked at refactoring the code to move the actual setting of the top level namespaces to the canonicalization code but not sure it gains us much. Will keep digging though.

LoneRifle:
Cool. I _think_ there is enough information in #167 to create a test case, which I would try to do this or next weekend, and then we'll see where we go from there.

bazzadp:
@LoneRifle I've gone ahead with that refactor after all. It makes the changes a good bit smaller and cleaner.  And also means I don't need to change c14nOptions which caused the problems before. Will leave you to review when you get a chance.

lib/signed-xml.js
```diff
@@ -449,59 +448,22 @@ SignedXml.prototype.validateReferences = function(doc) {
...
-          hasNonExcC14nTransform = true;
-          break;
-        }
-      }
-    
-      if(hasNonExcC14nTransform){
-        ref.ancestorNamespaces = findAncestorNs(doc, elemXpath);
-      }
+	  
+	  ref.ancestorNamespaces = findAncestorNs(doc, elemXpath);
```
LoneRifle: nitpick -
```suggestion
        ref.ancestorNamespaces = findAncestorNs(doc, elemXpath);
```

lib/signed-xml.js
```diff
@@ -612,9 +574,12 @@ SignedXml.prototype.loadReference = function(ref) {
       transforms.push(utils.findAttr(trans, "Algorithm").value)
     }
 
-    var inclusiveNamespaces = xpath.select("//*[local-name(.)='InclusiveNamespaces']", transformsNode);
+    var inclusiveNamespaces = utils.findChilds(trans, "InclusiveNamespaces")
     if (inclusiveNamespaces.length > 0) {
-      inclusiveNamespacesPrefixList = inclusiveNamespaces[0].getAttribute('PrefixList');
+      //Should really only be one prefix list, but maybe there's some circumstances where more than one to lets handle it
+      for (var i = 0; i<inclusiveNamespaces.length; i++) {
+        inclusiveNamespacesPrefixList = inclusiveNamespaces[i].getAttribute('PrefixList');
```
LoneRifle: Slightly doubtful of this - wouldn't the behaviour change such that you only get the last prefix list? Is that acceptable?

LoneRifle:
Turns out there wasn't quite enough information to construct a test case that displays the behaviour that was reported #167 . Without people being forthcoming with a self-contained test case of our own, I would like to proceed with merging this PR once my review is addressed

lib/signed-xml.js
bazzadp: Good spot. Fixed with new pull request. Not sure this will ever happen though, but fixed anyway.

LoneRifle:
Let me know if you want npm maintainer rights by replying with your npm login name. I'll publish this when next convenient.

mrcranky:
> To answer your question - #163 was originally filed to address ancestors used in canonicalization, but caused issue #167 and was hence backed out. I'm hoping that what you have done has properly fixed this, and will review in good time!

Just to follow up on this (I work with @gcharnock), #163 was raised because we see differing behaviour between xml-crypto and other libraries in relation to the specific signed SAML assertions we get from a third party. I've checked again with 1.4.0 and it definitely doesn't validate those assertions properly, even with this fix in. 1.0.2 works for us but obviously not for others. Right now we've forked the SAML libraries in use and pinned them to xml-crypto@1.0.2 but that's not ideal.

Unfortunately I can't share one of those assertions here as a test case because the third party in question is sensitive about that, if there is any advice anyone can give me for generating a sample XML doc that shows the problem? I'd like to get these assertions in as a test case to illustrate the issue so that subsequent fixes can try to address both #163 without breaking the other documents.

bazzadp:
Well this has since been merged and a test case has been added for my issue. So if you can fix it in the latest branch and it still passes the test cases then I’d be reasonably confident you shouldn’t break my usage and you can submit a pull request.

Of course without a test case you’re able to publicly share its difficult to help suggest a fix. Also even if you fix it and submit a pull request but don’t add a test case for your scenario, then no guarantee it won’t break in future.

If you look at the existing test cases, and coupled with what you know privately about your issue, then you might get some idea of how to create a test case?

PR #206, opened 2020-04-11
Add callback options to sign/verify asynchronously

troyfactor4:
-Hardware Security Module (HSM) signing/verification usually takes
 place over an async connection.
-Used callbacks instead of async/await for minimal code changes
 and backward compatibility

### DIFF
--- a/lib/signed-xml.js
+++ b/lib/signed-xml.js
@@ -333,28 +333,56 @@ SignedXml.defaultNsForPrefix = {
 
 SignedXml.findAncestorNs = findAncestorNs;
 
-SignedXml.prototype.checkSignature = function(xml) {
+SignedXml.prototype.checkSignature = function(xml, callback) {
+  if (callback != null && typeof callback !== 'function') {
+    throw new Error("Last paramater must be a callback function")
+  }
+
   this.validationErrors = []
   this.signedXml = xml
 
   if (!this.keyInfoProvider) {
-    throw new Error("cannot validate signature since no key info resolver was provided")
+    var err = new Error("cannot validate signature since no key info resolver was provided")
+    if (!callback){
+      throw err
+    }else{
+      callback(err)
+      return
+    }
   }
 
   this.signingKey = this.keyInfoProvider.getKey(this.keyInfo)
-  if (!this.signingKey) throw new Error("key info provider could not resolve key info " + this.keyInfo)
+  if (!this.signingKey) {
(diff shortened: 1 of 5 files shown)


lib/signed-xml.js
```diff
@@ -411,7 +439,7 @@ SignedXml.prototype.validateSignatureValue = function(doc) {
 SignedXml.prototype.calculateSignatureValue = function(doc) {
   var signedInfoCanon = this.getCanonSignedInfoXml(doc)
   var signer = this.findSignatureAlgorithm(this.signatureAlgorithm)
-  this.signatureValue = signer.getSignature(signedInfoCanon, this.signingKey)
+  return this.signatureValue = signer.getSignature(signedInfoCanon, this.signingKey)
```
LoneRifle, line 442: Having read the rest of this PR,  I have to ask - why break convention and potentially return a Promise here? Wouldn't it be more consistent to have a callback in the argument?

lib/signed-xml.js
```diff
@@ -333,28 +333,56 @@ SignedXml.defaultNsForPrefix = {
...
+    if (!callback){
+      return false;
+    }else{
+      callback(new Error('Could not validate references'))
+    }
   }
 
-  return true
+  var ret = this.validateSignatureValue(doc)
+  if (ret instanceof Promise){
```
LoneRifle, line 376: I wish to avoid Promises because I'm aware that xml-crypto is used in environments where getting the latest versions of node is a significant bureaucratic burden. To that end, I'm trying my best to support even versions of Node that are no longer maintained

troyfactor4:
Agreed with all change requests. Will fix and resubmit.

troyfactor4:
Do you know if arrow functions are safe to use? On line 813 of signed-xml.js after returning from a callback "this" was out of scope when using a standard anonymous function. Using an arrow function fixed the scoping issue.

The arrow function is only encountered in the "asynchronous flow" so I'm hoping it should not affect backwards compatibility. Maybe older version of NodeJS won't recognise that syntax though is what I'm worried about.

LoneRifle:
Arrow functions aren't supported, probably the cause of the build breakage. I'm afraid you will have to refactor that. Also, could you find and correct all instances of `if (!callback){` with appropriate whitespace?

LoneRifle:
Made a PR to your PR to cleanup the work at troyfactor4/xml-crypto#1

troyfactor4:
Wow, thanks! Appreciate it. Looks great!

PR #207, opened 2020-04-12
Test suites of other projects (mocha) that include v1.5.0 fail

troyfactor4:
Error message: "Error: global leak detected: existingPrefixes"
Test was run using project 'samlp' from npmjs.org with xml-crypto updated to 1.5.0

Steps to reproduce:

1. git clone https://github.com/auth0/node-samlp.git
2. Modify package.json and update xml-crypto to 1.5.0
3. npm install
4. npm test


troyfactor4:
I can make these two separate PR if needed.

PR #209, opened 2020-04-13
Async response for built in algo sign/verify

troyfactor4:
Easier async integration in other projects when using built in sign/verify algorithms

### DIFF
--- a/lib/signed-xml.js
+++ b/lib/signed-xml.js
@@ -85,22 +85,24 @@ function RSASHA1() {
   * Sign the given string using the given key
   *
   */
-  this.getSignature = function(signedInfo, signingKey) {
+  this.getSignature = function(signedInfo, signingKey, callback) {
     var signer = crypto.createSign("RSA-SHA1")
     signer.update(signedInfo)
     var res = signer.sign(signingKey, 'base64')
-    return res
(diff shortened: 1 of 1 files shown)


troyfactor4:
Apologies for the many PR's. This should be the last one.

LoneRifle:
@troyfactor4 - given that this library is passively maintained, I thought you might want to know that there are a couple of other libraries that provides XML Digital Signatures:
[xmldsigjs](https://github.com/PeculiarVentures/xmldsigjs) - support for XMLDSIG
[xadesjs](https://github.com/PeculiarVentures/xadesjs) - support for XAdES and XMLDSIG built on top of xmldsigjs

The READMES seem to indicate that the libraries are not production-ready, but they are already in development ofr 3 years and have v2.x releases, so perhaps it might be good to discuss whether you should use their libraries instead with @microshine

cjbarth:
@LoneRifle is this an official recommendation? That is to say, is this library no longer adding value to the NPM ecosystem?

LoneRifle:
That's from long ago, before you came along and did active management again!

cjbarth:
I understand that. I'm just wondering if other libraries do what this library does, then let's work together and avoid duplicate effort.

Issue #195, opened 2019-10-03
Unnecessary new line and whitespace normalization in canonicalization alg


nicosabena:
Hi guys.
After troubleshooting some signature validation errors on SAML responses (specifically, unmatching digests), I think I found a couple of things currently done on the algorithms that should not be done. In particular:

- Normalization of line endings for element text:
https://github.com/yaronn/xml-crypto/blob/40bd9e8f8787c9c996572fe5937572506b9e43d1/lib/utils.js#L69
- Normalization of whitespace for attributes:
https://github.com/yaronn/xml-crypto/blob/40bd9e8f8787c9c996572fe5937572506b9e43d1/lib/utils.js#L58

In https://www.w3.org/TR/xml-c14n11/#DataModel it says that "the XML processor performs the following tasks in order":

1. normalize line feeds
2. normalize attribute values
3. replace CDATA sections with their character content
4. resolve character and parsed entity references

Since the resolve character and parsed entity references step comes after the line feed normalization, this means that an input like this:

```
<bar>\r\n
  <foo>&#xD;\n</foo>\r
</bar>\r\n
```
(new lines added for clarity)

will result in this after the parser is done:

```
<bar>\n
  <foo>\r\n</foo>\n
</bar>\n
```

The `&#xD;` entity that was encoded in `foo` should be left alone in the contents, and should be replaced by the `&xD;` entity when calculating the digest. The code, by normalizing the line feeds again as part of its processing, is removing that `\r` and altering the results.

A similar situation happens with attributes and whitespace normalization. The canonicalization process should not alter white space that was already processed by the parser. E.g.
```
<bar weirddata="&#x9;&#xA;&#xD" />
```
becomes this after the parser:

```
<bar weirddata="\t\n\r" />
```

That should be re-coded to their entity references for canonicalization, but the code is currently transforming the above to a single space character.

This is further demonstrated in the examples provided at https://www.w3.org/TR/xml-c14n11/#Example-Chars

PR #211, opened 2020-05-13
Don't encode windows line-breaks with special characters.

thenengah:
Getting "Invalid Signature". The line breaks are getting encoded as special characters which botch the signature checking.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -65,7 +65,7 @@ function encodeSpecialCharactersInAttribute(attributeValue){
 
 function encodeSpecialCharactersInText(text){
     return text
-	      .replace(/\r\n/g, '\n')  // Line ending normalization (Note: this should normally be done by the xml parser). See: https://www.w3.org/TR/xml/#sec-line-ends
+	.replace(/\r\n/g, '\n')  // Line ending normalization (Note: this should normally be done by the xml parser). See: https://www.w3.org/TR/xml/#sec-line-ends
         .replace(/([&<>\r])/g, function(str, item){
             // Special character normalization. See:
             // - https://www.w3.org/TR/xml-c14n#ProcessingModel (Text Nodes)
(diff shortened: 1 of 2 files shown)


thenengah:
Made the update less restrictive to only replace windows line breaks. I also updated the windows line-break test to expect the line-break to change as encoding it is what was producing the invalid signature.

LoneRifle:
This is intended, per the issue raised in #195 and addressed in #196. I've cc'd @nicosabena to explain further.

nicosabena:
Hi @thenengah ! 

Basically line end normalization is the job of the XML Processor, before parsing (https://www.w3.org/TR/xml/#sec-line-ends). If there are Windows line breaks after the parser runs, those should be treated as the result of parser-resolved character references. E.g.

Raw input to the parser:
```
<foo>&#xD;\n</foo>
```

The parser will return `\r\n` as the data for `foo`'s text element. This shouldn't be altered by the canonicalization algorithm.

[xmldoc](https://github.com/xmldom/xmldom) does not normalize line endings automatically so, If anything, you can (should) normalize the line endings when reading the source:

```
function normalizeLineEndings(string) {
  return string.replace(/\r\n?/g, '\n');
}
var xml = await readFileFromSource();
var doc = xmldom.DOMParser().parseFromString(normalizeLineEndings(xml));
```

Can you give that a try and see if it solves the signature validation?

thenengah:
Ok, thanks for the help! 

We already implemented this:

```
  const xml = (Buffer.from(req.body.SAMLResponse, 'base64').toString()).replace(/\r\n?/g, '\n')
  req.body.SAMLResponse = Buffer.from(xml.toString()).toString('base64')
```

We had to do this to get it working with passport-saml@1.3.3. We have to schedule testing with some affected customers to confirm this actually works. Will confirm here later if so.

PR #220, opened 2021-01-29
Don't pull the example folder into the module build

dosullivan557:


dosullivan557:
Hi @gcharnock @bergie @locnguyen @cava23 - could you review this pr please.

I install your module in an application, it copies over .pem and .key files that are in your tests. This means that when we put our application through security scans, it fails because it finds these files and classifies them as sensitive data. What I'm asking is whether you can not include these in the bundled version of the application.

To resolve it, you can have a section in the package.json that doesn't pull the test files over to the build of the module when it is pulled in.

dosullivan557:
Resolves #221

cava23:
I thought the package-lock.json should never be hand-modded. Is there a corresponding change to package.json you're trying to make? I don't understand the intended change based on the diff.

dosullivan557:
The package lock has only been upversioned and it wasn't done manually, I was done using the nom version patch command.

dosullivan557:
@cava23  if you look at the latest changes, there is a files section in the package.json that specifies which folders and files need putting into the package when installed.

cava23:
Can you remove whitespace changes so the diff is legible?

dosullivan557:
@cava23 Sure thing - removed whitespace

cava23:
Sorry, looked at this a little closer and it appears this is already being attempted in the `.npmignore` file. However, I believe the trailing slash is missing. I think the right change here is to update `.npmignore` to have `test/`. See if that solves your problem. I'd rather use that than add the `files` prop to package.json since that could have a greater impact than intended.

cava23:
You can test this with `npm pack`: https://docs.npmjs.com/cli/v6/using-npm/developers#testing-whether-your-npmignore-or-files-config-works

dosullivan557:
Moved to .gitignore, and added example in there, which is causing an issue too. @cava23

cava23:
Nice, I assume you meant .npmignore instead .gitignore. And this is now working as expected?

dosullivan557:
Yeah sorry I mean .npmignore.

And yep all working as expected

cava23 (approved):
I would like to get approval from one other contributor as I've only made a small change to this project.

dosullivan557:
Sounds fair enough - cheers!

dosullivan557:
Hey @yaronn - are you able to review this pull request, please.

LoneRifle (approved):
lgtm

PR #242, opened 2022-03-17
Correct behavior for XML canonicalization with namespaces and nested elements

mthadley:
Hey there! I wanted to start by saying thank you for the hard work maintaining this package.

I've been running into problems related to mismatched signature digests when validating a signed XML document, and I've narrowed the mismatch down to differences in how this library and another canonicalize the document. While the real world example involves SAML response XML, I've tried to instead demonstrate the behavior with a minimal failing test.

For the `http://www.w3.org/TR/2001/REC-xml-c14n-20010315` method, I noticed that `xml-crypto` seems to be repeating namespaces that are already defined on an ancestor node.

If we start with this document...

```xml
<root xmlns:aaa='bbb'>
  <child1>
    <child2>
      <child3 aaa:foo='bar'>
      </child3>
    </child2>
  </child1>
</root>
```

...and like in the test, canonicalize the subset of the document with `child1`, we get...

```xml
<child1 xmlns:aaa="bbb">
  <child2>
    <child3 xmlns:aaa="bbb" aaa:foo="bar">
    </child3>
  </child2>
</child1>
```

Notice that `xmlns:aaa` gets pushed down from the `root` to `child1`, which I expect, but then the namespace is also repeated on `child3`.

Is this expected behavior? It seems incorrect, due to the duplication of `xmlns:aaa`.

### DIFF
--- a/lib/c14n-canonicalization.js
+++ b/lib/c14n-canonicalization.js
@@ -208,7 +208,13 @@ C14nCanonicalization.prototype.process = function(node, options) {
   var defaultNsForPrefix = options.defaultNsForPrefix || {};
   var ancestorNamespaces = options.ancestorNamespaces || [];
 
-  var res = this.processInner(node, [], defaultNs, defaultNsForPrefix, ancestorNamespaces);
+  var prefixesInScope = [];
+  for (var i = 0; i < ancestorNamespaces.length; i++) {
+    prefixesInScope.push(ancestorNamespaces[i].prefix)
+  }
+
+  console.log({prefixesInScope})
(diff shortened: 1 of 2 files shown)


mthadley:
In addition to the failing test, I've pushed up what I think may be a fix, in [27ca996](https://github.com/yaronn/xml-crypto/pull/242/commits/27ca996f24acd5e37a54444d4d9cd0fbbf41bed5).

cjbarth (approved):

LoneRifle (approved):
lgtm, taking reference from @cjbarth and node-saml/node-saml#36

Issue #243, opened 2022-05-08
Allow FileKeyInfo to accept a string for the cert


cjbarth:
Currently `FileKeyInfo` only accepts a file path for construction. The means that the cert must be on a file system somewhere. In our security context, the certs and keys aren't on disk, they are in memory, so ideally we'd like to construct `FileKeyInfo` using a string that is the cert, not a path.

Would you be open to a PR that changed [this](https://github.com/yaronn/xml-crypto/blob/63acff9fec59e32996e6a13d79ff69e7e51a9f44/lib/signed-xml.js#L17-L29) code to accept a string that is either a file path or a cert? The basic switch would be that if the string isn't a valid path, then treat the string as if it were a cert and proceed as normal.

cjbarth:
I also note that the parameter `key` isn't used in [`getKeyInfo()`](https://github.com/yaronn/xml-crypto/blob/63acff9fec59e32996e6a13d79ff69e7e51a9f44/lib/signed-xml.js#L20-L24), which seems like it might be a bug. Thoughts?

cjbarth:
Finally, in this same area of code, it doesn't seem that there is support for multiple certs as seems is supported by [SAML](https://lists.w3.org/Archives/Public/public-xmlsec-maintwg/2007Jun/att-0068/DNAMES_Section4.4.4.htm). Is that something that could be added?

cjbarth:
Also, there appears to be no support for multiple `<x509Data>` elements, per the spec here: https://www.w3.org/TR/xmldsig-core1/#sec-X509Data.

LoneRifle:
@cjbarth - FileKeyInfo is a convenience implementation of a KeyInfoProvider interface, which exposes just two methods, `getKeyInfo()` and `getKey()`. As the name suggests, the class is backed by a file handle.

Is there any reason why you don't wish to maintain your own implementation of KeyInfoProvider?

cjbarth:
> Is there any reason why you don't wish to maintain your own implementation of KeyInfoProvider?

There are at least two reasons:

1. We want to allow the community to benefit from this work, so we want it to be upstream.
2. We want to make sure that the work is done correctly and in a maintainable way, so we want to discuss the changes in the correct repo.

cjbarth:
> FileKeyInfo is a convenience implementation of a KeyInfoProvider interface, which exposes just two methods, `getKeyInfo()` and `getKey()`. As the name suggests, the class is backed by a file handle.

[...]

PR #249, opened 2022-07-18
Updated getKeyInfo function with actual implementation

ganesha289:
Hello @cjbarth, @LoneRifle, @yaronn,
I have added implementation for getKeyInfo function, please check if it looks good, then I'll proceed with unit tests.   

I am getting confused with below line of codes,

For FileKeyInfo, I'm assuming it accepts file which will be public X509 certificate and this.getKeyInfo function will return the X509Data. **But, what is the use of this.getKey()??** I'm assuming it will return public certificate. But, in some places it used as privateKey, as shown in below statements,

In node-saml->src->xml.ts->signXml, signingKey gets the privateKey value assigned, as shown below,
image

However, in xml-crypto->lib->signed-sml.js, signingKey gets the value from KeyInfoProvider.getKey(this.keyInfo) which looks to be public certificate..
image

Can you please help me to clear this..

### DIFF
--- a/lib/signed-xml.js
+++ b/lib/signed-xml.js
@@ -870,7 +870,7 @@ SignedXml.prototype.getKeyInfo = function(prefix) {
 
   if (this.keyInfoProvider) {
     res += "<" + currentPrefix + "KeyInfo>"
-    res += this.keyInfoProvider.getKeyInfo(this.signingCert, prefix)
+    res += this.keyInfoProvider.getKeyInfo(this.signingCert || this.signingKey, prefix)
     res += "</" + currentPrefix + "KeyInfo>"
   }
   return res
--- a/lib/signed-xml.js
+++ b/lib/signed-xml.js
@@ -18,12 +18,13 @@ function FileKeyInfo(file) {
   this.file = file
 
   this.getKeyInfo = function(key, prefix) {
-    let currentPrefix = prefix || ''
+    var currentPrefix = prefix || ''
     currentPrefix = currentPrefix ? currentPrefix + ':' : currentPrefix
-    let signingCert = '';
+    var signingCert = '';
     if (key) {
-      let certArray = Array.isArray(key) ? key : [key];
-      for(let cert in certArray) {
+      var certArray = [].concat(key);
+      for(var i = 0; i < certArray.length; ++i) {
+        var cert = certArray[i]
         signingCert += "<" + currentPrefix + "X509Certificate>" + cert + "</" + currentPrefix + "X509Certificate>"
       }
     }
--- a/lib/signed-xml.js
+++ b/lib/signed-xml.js
@@ -18,9 +18,16 @@ function FileKeyInfo(file) {
   this.file = file
 
   this.getKeyInfo = function(key, prefix) {
-    prefix = prefix || ''
-    prefix = prefix ? prefix + ':' : prefix
-    return "<" + prefix + "X509Data></" + prefix + "X509Data>"
+    let currentPrefix = prefix || ''
+    currentPrefix = currentPrefix ? currentPrefix + ':' : currentPrefix
+    let signingCert = '';
+    if (key) {
+      let certArray = Array.isArray(key) ? key : [key];
+      for(let cert in certArray) {
+        signingCert += "<" + currentPrefix + "X509Certificate>" + cert + "</" + currentPrefix + "X509Certificate>"
+      }
+    }
+    return "<" + currentPrefix + "X509Data>" + signingCert + "</" + currentPrefix + "X509Data>"
   }
 
   this.getKey = function(keyInfo) {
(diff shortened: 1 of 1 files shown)


cjbarth:
@ganesha289 , I'm not sure what the confusion is. Signing should be done with a private key and verification of a signature is done with a public key. The code that you reference in `signed-xml.js` is part of the function called `checkSignature`, so it would make sense that it would use a public cert. As for `node-saml`, that looks like it is correctly using a private key to sign an XML document.

lib/signed-xml.js
```diff
@@ -18,9 +18,16 @@ function FileKeyInfo(file) {
...
+    let currentPrefix = prefix || ''
+    currentPrefix = currentPrefix ? currentPrefix + ':' : currentPrefix
+    let signingCert = '';
+    if (key) {
+      let certArray = Array.isArray(key) ? key : [key];
+      for(let cert in certArray) {
+        signingCert += "<" + currentPrefix + "X509Certificate>" + cert + "</" + currentPrefix + "X509Certificate>"
+      }
+    }
+    return "<" + currentPrefix + "X509Data>" + signingCert + "</" + currentPrefix + "X509Data>"
```
LoneRifle, line 30: We're trying to maintain backward compatibility with Node 4 here, since enterprise environments might be extremely slow in moving. Could we do..
```suggestion
    var currentPrefix = prefix || ''
    currentPrefix = currentPrefix ? currentPrefix + ':' : currentPrefix
    var signingCert = '';
    if (key) {
      var certArray = [].concat(key);
      for(var i = 0; i < certArray.length; ++i) {
        var cert = certArray[i]
        signingCert += "<" + currentPrefix + "X509Certificate>" + cert + "</" + currentPrefix + "X509Certificate>"
      }
    }
    return "<" + currentPrefix + "X509Data>" + signingCert + "</" + currentPrefix + "X509Data>"
```

lib/signed-xml.js
```diff
@@ -861,7 +869,7 @@ SignedXml.prototype.getKeyInfo = function(prefix) {
 
   if (this.keyInfoProvider) {
     res += "<" + currentPrefix + "KeyInfo>"
-    res += this.keyInfoProvider.getKeyInfo(this.signingKey, prefix)
+    res += this.keyInfoProvider.getKeyInfo(this.signingCert, prefix)
```
LoneRifle, line 872: Potential breaking change, as existing users will be relying on `signingKey` to generate the key info. What are you trying to do here?

LoneRifle (changes requested):

lib/signed-xml.js
```diff
@@ -861,7 +870,7 @@ SignedXml.prototype.getKeyInfo = function(prefix) {
...
```
LoneRifle, line 873: potentially breaking change, since existing users may be providing `signingKey`. Is there a better way?

cjbarth: There might be, but right now none of the code in the code-base actually uses that first argument, so what does it matter?

There is only one concrete implementation of `KeyInfoProvider`, which is `FileKeyInfo`, and that doesn't use the first argument at all.

https://github.com/yaronn/xml-crypto/blob/3deacfcd519b5ba97ee67ca5f0046d8cc4a7a78c/lib/signed-xml.js#L20-L24

cjbarth: Granted, this PR does change that, but we need to establish how things are supposed to work first, and then we can consider a change. I'm still considering your advice and what the README says about rolling our own instead of enhancing `FileKeyInfo`. See #243 .

LoneRifle: @cjbarth  - So I agree with your arguments in #243 to add an upstream in-memory `KeyInfoProvider`. What I think we can do to provide some semblance of backward compatibility is to supply `signingKey` as an argument if present, otherwise falling back on `signingCert`. Something like...
```suggestion
    res += this.keyInfoProvider.getKeyInfo(this.signingKey || this.signingCert, prefix)
```

wdyt? If you're happy with this I'm happy for this to rapidly approve, merge and release this PR

cjbarth:
@ganesha289 I think we have a general idea of how we'd like to move forward. Since @LoneRifle and I are pretty much the only ones left doing anything that looks like maintenance on this project, I think the three of us can figure out the correct pattern here, implement it, and close this out. I know `node-saml` will benefit from it, and I expect others will too, especially if the pattern is clearly established.

lib/signed-xml.js
djaqua: @LoneRifle I like where you're going with that compromise, but wouldn't it make more sense to "prefer the new way and default to the old way?" (e.g. `this.signingCert || this.signingKey`)

LoneRifle: Good point!

LoneRifle (approved):
lgtm

PR #273, opened 2023-01-17
make FileKeyInfo extensible for compatibility with TypeScript

### LINKED ISSUE #272: Issue with TypeScript and FileKeyInfo
Even with @types/xml-crypto, I am unable to _usefully_ extend FileKeyInfo. The syntax checks out and TypeScript compiles, but the signature does not honor the methods `getKey` and `getKeyInfo`, but proceeds to use the methods as currently defined on FileKeyInfo.

Since TypeScript is strongly typed, I am unable to assign anything to the `keyInfoProvider` property of the signature object besides instances of KeyInfoProvider and its subclasses. 

I have figured out that this happens because both methods are properties of the instance of FileKeyInfo itself, and not its prototype. 

The solution is simple, but perhaps you know of a reason not to proceed (for backward's compatibility reasons etc etc). Change:
```
function FileKeyInfo(file) {
  this.file = file;

  this.getKeyInfo = function (key, prefix) {
    prefix = prefix || "";
    prefix = prefix ? prefix + ":" : prefix;
    return "<" + prefix + "X509Data></" + prefix + "X509Data>";
  };

  this.getKey = function (keyInfo) {
    return fs.readFileSync(this.file);
  };
}
```

To this:
```
function FileKeyInfo(file) {
  this.file = file;
}

FileKeyInfo.prototype.getKeyInfo = function (key, prefix) {
  prefix = prefix || "";
  prefix = prefix ? prefix + ":" : prefix;
  return "<" + prefix + "X509Data></" + prefix + "X509Data>";
};

FileKeyInfo.prototype.getKey = function (keyInfo) {
  return fs.readFileSync(this.file);
}
```

I tested this locally and it passed all your existing unit tests and it also solved my issue. I would be happy to create a PR and submit for review :)

djaqua:
The type definitions for this awesome library require that the value of `SignedXML.keyInfoProvider` be an instance of `FileKeyInfo`, but the methods on FileKeyInfo belong to the instance itself and not the prototype. This issue is only strictly a problem for TypeScript, but this change has a beneficial affect (in terms of reuse) on NodeJS clients, as well, and this is a non-breaking change. 

Also adding some documentation to the methods to clear up some of the confusion. Feedback is welcome!

Closes #272

### DIFF
--- a/lib/file-key-info.js
+++ b/lib/file-key-info.js
@@ -1,0 +1,17 @@
+var StringKeyInfo = require("./string-key-info")
+  , fs = require("fs")
+
+/**
+ * A key info provider implementation
+ * 
+ * @param {string} file path to public certificate 
+ */
+function FileKeyInfo(file) {
+  var key = fs.readFileSync(file)
+  StringKeyInfo.apply(this, [key])
+}
+
+FileKeyInfo.prototype = StringKeyInfo.prototype
+FileKeyInfo.prototype.constructor = FileKeyInfo
+
+module.exports = FileKeyInfo
--- a/lib/signed-xml.js
+++ b/lib/signed-xml.js
@@ -4,51 +4,13 @@ var xpath = require('xpath')
   , c14n = require('./c14n-canonicalization')
   , execC14n = require('./exclusive-canonicalization')
   , EnvelopedSignature = require('./enveloped-signature').EnvelopedSignature
+  , StringKeyInfo = require('./string-key-info')
+  , FileKeyInfo = require('./file-key-info')
   , crypto = require('crypto')
-  , fs = require('fs')
 
 exports.SignedXml = SignedXml
+exports.StringKeyInfo = StringKeyInfo
 exports.FileKeyInfo = FileKeyInfo
-
-/**
- * A key info provider implementation
- *
- */
-function FileKeyInfo(file) {
-  this.file = file
-}
-
-
-/**
- * Builds the contents of a KeyInfo element as an XML string.
- *
- * Currently, this returns exactly one empty X509Data element
- * (e.g. "<X509Data></X509Data>"). The resultant X509Data element will be
- * prefaced with a namespace alias if a value for the prefix argument
- * is provided. In example, if the value of the prefix argument is 'foo', then
- * the resultant XML string will be "<foo:X509Element></foo:X509Element>"
- *
- * @param key (not used) the signing/private key as a string
- * @param prefix an optional namespace alias to be used for the generated XML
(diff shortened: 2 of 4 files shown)


cjbarth:
Should this be a more generic class such that other types of `KeyInfo` can be used, like `StringKeyInfo`? I'm thinking of the work done [here](https://github.com/yaronn/xml-crypto/pull/249). It would be nice if all this related code could be coordinated.

djaqua:
I'm in favor of that. About to push a change to the PR ... let me know what you think

djaqua:
@cjbarth I just pushed up changes to my branch, let me know what you think!

cjbarth:
Those are some nice changes. I need a little more time to compare them to #249 and related code over at `node-saml`: https://github.com/node-saml/node-saml/pull/36

These issues have been open a while and I would really appreciate your help to wrangle them all together and get the functionality landed. Any help or comments you provide would be most welcome.

djaqua:
@cjbarth & @LoneRifle - I do see why these issues are all related but I think the reality is that if you want to wrangle all the outstanding issues, then we're going to need a new major version number and put some breaking changes into it (that's fair play for a major version change). I'm putting together a proposal to address those issues for the time being and will share once I have something useful to work with.

Either way, I propose merging this PR as a quick victory for the current major version. It doesn't block any solutions for the other issues, and solves the extensibility issue as well as providing a string-based implementation (issue #243 ). **At its best**, it can be used as a valid step towards addressing the other issues if we choose not to pursue the breaking-changes route. At it's worst, it's a tidy solution for the current version if we do choose to pursue the breaking-changes route.

LoneRifle:
I don't mind. I think we also agreed that we want to do a major breaking version anyway, since xml-crypto has moved to @node-saml. The main blocker would be the lack of types from DefinitelyTyped, but this can be easily addressed by moving the types into this repo.

I realize there is a fair bit of yak-shaving to do, so @djaqua if we need to move quickly, let me know, and I'll strive to find some time to look into this.

cc @cjbarth

djaqua:
@LoneRifle I've actually been in contact with the maintainer of the maintainer of the Definitely Typed type definitions for this project (there's a bug I volunteered to fix) and once I joined you and @cjbarth on this project I actually asked him to hold off on the fix until the dust settles here.

I don't mind being a co-maintainer or liaison between this project and the type definitions project - I would be glad to help coordinate the maintenance and ensure continuity between the projects 🤷‍♂️

I thought I was the only person still using the expression "yak-shaving" lol

LoneRifle (approved):
lgtm

PR #277, opened 2023-02-26
feat: add type declaration

LoneRifle:
- port type declaration from @DefinitelyTyped, and reference it from package.json
- add definitions for KeyInfo, StringKeyInfo, tweak SignedXml to take a KeyInfo for `keyInfoProvider` instead of FileKeyInfo

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,98 @@
+// Type definitions for @node-saml/xml-crypto
+// Project: https://github.com/node-saml/xml-crypto#readme
+// Original definitions by: Eric Heikes <https://github.com/eheikes>
+//                          Max Chehab <https://github.com/maxchehab>
+
+/// <reference types="node" />
+
+import { SelectedValue } from 'xpath';
+
+export class HashAlgorithm {
+    getAlgorithmName(): string;
+    getHash(xml: string): string;
+}
+
+export interface Reference {
+    xpath: string;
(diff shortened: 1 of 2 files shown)


cjbarth:
I am on a special project for about 4 more weeks, so I'll be delayed in anything GitHub-related until after then.

index.d.ts
```diff
@@ -0,0 +1,98 @@
...
+        }
+    ): void;
+    getOriginalXmlWithIds(): string;
+    getSignatureXml(): string;
+    getSignedXml(): string;
+    loadSignature(signatureNode: string | Node): void;
+}
+
+export interface KeyInfo {
+    getKey(keyInfo?: Node): Buffer;
```
djaqua, line 80: Ha! I brought this issue up to the DT maintainer: getKey should be accepting an array of Nodes, not a single Node object

I will link the issue and explanation here in a moment

djaqua (changes requested):

djaqua:
Save for the one comment I left, this looks great!

index.d.ts
LoneRifle: So something like...
```suggestion
    getKey(keyInfo?: null | Node[]): Buffer;
```
Per that discussion?

djaqua (approved):
Awesome 🙌

cjbarth (approved):

Issue #228, opened 2021-04-12
How to add InclusiveNamespaces in Signature


lostvicking:
Hello and thank you for maintaining this very useful project! I am using `xml-crypto` indirectly through a project called `smalify`. We have a requirement to have an `ec:InclusiveNamespaces` element in the signature of our SAML Response. I cannot seem to figure out how one adds InclusiceNamespaces to a signature and would be very thankful for some guidance on how to get that right. 

```xml
<ds:SignedInfo>
<ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
<ds:SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/>
<ds:Reference URI="#_1e1ed6a7c84fb5787ff829b0ba7a23b4">
<ds:Transforms>
<ds:Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/>
<ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
<ec:InclusiveNamespaces xmlns:ec="http://www.w3.org/2001/10/xml-exc-c14n#" PrefixList="ds saml xs xsi" />
</ds:Transforms>
```

Issue #189, opened 2019-06-25
Documentation for adding InclusiveNamespaces and PrefixList to CanonicalizationMethod and Transforms


crabey:
Hello

I’m having trouble getting InclusiveNamespaces to work like intended. It would be great to have an example in the library documentation that shows how to achieve this.

I want my XML SignedInfo part to look like this:
```
<ds:SignedInfo>
  <ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#">
    <ec:InclusiveNamespaces xmlns:ec="http://www.w3.org/2001/10/xml-exc-c14n#" PrefixList="soapenv urn urn1 urn2"/>
  </ds:CanonicalizationMethod>
  <ds:SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/>
  <ds:Reference URI="#TS-94B652AC5EA054E08B156034518270218">
    <ds:Transforms>
      <ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#">
        <ec:InclusiveNamespaces xmlns:ec="http://www.w3.org/2001/10/xml-exc-c14n#" PrefixList="wsse soapenv urn urn1 urn2"/>
      </ds:Transform>
    </ds:Transforms>
    <ds:DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/>
    <ds:DigestValue>h5Evftf429SeUkW0UMnMt2EregM=</ds:DigestValue>
  </ds:Reference>
  <ds:Reference URI="#id-94B6F97C5E5232E08B156034518272522">
    <ds:Transforms>
      <ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#">
        <ec:InclusiveNamespaces xmlns:ec="http://www.w3.org/2001/10/xml-exc-c14n#" PrefixList="urn urn1 urn2"/>
      </ds:Transform>
    </ds:Transforms>
    <ds:DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/>
    <ds:DigestValue>IgGy8/m034pkLmcDBBlzQLHl6rA=</ds:DigestValue>
  </ds:Reference>
</ds:SignedInfo>
```

This is what I’m currently getting:
```
<ds:SignedInfo>
  <ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
[...]

Issue #138, opened 2017-07-17
InclusiveNamespaces how to include inclusivenamespace


SSANSH:
Hi, How I can generate transform definition like with InclusiveNamespaces.

   ```
<ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#">
      <InclusiveNamespaces xmlns="http://www.w3.org/2001/10/xml-exc-c14n#" PrefixList="oas v1 v11"/>
    </ds:Transform>
```

I checked the function SignedXml.prototype.createReferences but I don't see the way to include this definition.

      res += "<" + prefix + "Transforms>"
      for (var t in ref.transforms) {
        if (!ref.transforms.hasOwnProperty(t)) continue;

        var trans = ref.transforms[t]
        var transform = this.findCanonicalizationAlgorithm(trans)
        res += "<" + prefix + "Transform Algorithm=\"" + transform.getAlgorithmName() + "\"/>";
      }

Can you explain how to do and if its doable?

Thanks.

tngan:
https://github.com/yaronn/xml-crypto/issues/94

@SSANSH checkout this one and you could find out the workaround as well

shunkica:
@SSANSH did you ever figure it out?

I am parsing the signed document back to a xml doc and then manually adding the InclusiveNamespaces to ds:CanonicalizationMethod and ds:Transform but this is a very hacky approach.

Issue #60, opened 2015-06-24
Include InclusiveNamespaces in transform tag


cmartin81:
Hi
I need to define a reference when signing my xml with a html tag called: InclusiveNamespaces.
Are there any solution for this?

This is what I need to have inside my reference tag:
<Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#">
   <InclusiveNamespaces  xmlns="http://www.w3.org/2001/10/xml-exc-c14n#" PrefixList="#default samlp saml ds xs xsi"/>
</Transform>

yaronn:
Hi @cmartin81 

This is not supported yet but some of the infrustructure so it might not be complex to implement if you want to try. Check this part in signed-xml.js:

https://github.com/auth0/xml-crypto/blob/be60701831079964ca56dbbd87a263af689fa509/lib/signed-xml.js#L416

you need two changes:

1. the getCannonXml() method can get a third parameter with the namespaces you want in the PrefixList (see how to call it here https://github.com/auth0/xml-crypto/blob/be60701831079964ca56dbbd87a263af689fa509/lib/signed-xml.js#L246)

2. In the same place where you added the parameter, see how the "transform" xml tag is built. you need to add it the "InclusiveNamespaces" child tag as in your example (using simple string concat). 

You can try to do it hard coded with your required namespaces so it will be very fast to check if it works.

Thanks,
Yaron

zygimantus:
So, InclusiveNamespaces are not supported yet?

PR #284, opened 2023-05-24
Use inclusiveNamespacesPrefixList to generate InclusiveNamespaces

shunkica:
Currently, the `inclusiveNamespacesPrefixList` property on the Reference interface is only used during loading of references.
This change would allow the use of the same property during the creation of references to add a `InclusiveNamespaces` element.

Example:

```
sig.addReference(
    "//*[local-name(.)='Timestamp']",
    ["http://www.w3.org/2001/10/xml-exc-c14n#"],
    'http://www.w3.org/2000/09/xmldsig#sha1',
    '','',
    'wsse soapenc soapenv xsd xsi');
```

Result:

```
<ds:Reference URI="#TS-b3f4bfa7-fe3b-4ee1-aa8c-289ee7de442b">
  <ds:Transforms>
       <InclusiveNamespaces PrefixList="wsse soapenc soapenv xsd xsi"
          xmlns="http://www.w3.org/2001/10/xml-exc-c14n#"/>
     </ds:Transform>
   </ds:Transforms>
   <ds:DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/>
   <ds:DigestValue>DGrj1zr8K0OsBojzMSMVowotX60=</ds:DigestValue>
</ds:Reference>
```
    
Added `inclusiveNamespacesPrefixList` property to SignedXML options, which if present generates InclusiveNamespaces element inside CanonicalizationMethod.

Example:

```
const sig = new SignedXml(null, {idAttribute: 'Id', inclusiveNamespacesPrefixList: 'soapenc soapenv xsd xsi'});
```

Result:

```
<ds:Signature Id="SIG-0271048c-6ea7-4e9d-85e2-1819c3ce3848" xmlns:ds="http://www.w3.org/2000/09/xmldsig#">
  <ds:SignedInfo>
    <ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#">
      <InclusiveNamespaces PrefixList="soapenc soapenv xsd xsi"
        xmlns="http://www.w3.org/2001/10/xml-exc-c14n#"/>
      </ds:CanonicalizationMethod>
```

### DIFF
--- a/lib/signed-xml.js
+++ b/lib/signed-xml.js
@@ -286,6 +286,7 @@ function SignedXml(idMode, options) {
   this.signatureAlgorithm = this.options.signatureAlgorithm || "http://www.w3.org/2000/09/xmldsig#rsa-sha1";
   this.keyInfoProvider = null
   this.canonicalizationAlgorithm = this.options.canonicalizationAlgorithm || "http://www.w3.org/2001/10/xml-exc-c14n#"
+  this.inclusiveNamespacesPrefixList = this.options.inclusiveNamespacesPrefixList || ""
   this.signedXml = ""
   this.signatureXml = ""
   this.signatureNode = null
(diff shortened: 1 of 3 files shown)


lib/signed-xml.js
```diff
@@ -891,7 +892,14 @@ SignedXml.prototype.createReferences = function(doc, prefix) {
 
         var trans = ref.transforms[t]
         var transform = this.findCanonicalizationAlgorithm(trans)
-        res += "<" + prefix + "Transform Algorithm=\"" + transform.getAlgorithmName() + "\" />"
+        res += "<" + prefix + "Transform Algorithm=\"" + transform.getAlgorithmName() + "\""
+        if (ref.inclusiveNamespacesPrefixList && ref.inclusiveNamespacesPrefixList.length > 0) {
```
cjbarth, line 896: You check to see if `.length > 0`, but I think you mean to test `.length === 1` because I see no loop to handle anything other than a length of 1.

lib/signed-xml.js
```diff
@@ -984,8 +992,15 @@ SignedXml.prototype.createSignedInfo = function(doc, prefix) {
   currentPrefix = currentPrefix ? currentPrefix + ':' : currentPrefix
 
   var res = "<" + currentPrefix + "SignedInfo>"
-  res += "<" + currentPrefix + "CanonicalizationMethod Algorithm=\"" + transform.getAlgorithmName() + "\" />" +
-          "<" + currentPrefix + "SignatureMethod Algorithm=\"" + algo.getAlgorithmName() + "\" />"
+  res += "<" + currentPrefix + "CanonicalizationMethod Algorithm=\"" + transform.getAlgorithmName() + "\""
+  if (this.inclusiveNamespacesPrefixList.length > 0) {
```
cjbarth, line 996: See above comments about length.

cjbarth (changes requested):

lib/signed-xml.js
shunkica: The property `inclusiveNamespacesPrefixList` is a string, not an array. But you are correct, the length check is not needed at all.

cjbarth (approved):

shunkica:
This pull request closes #228, #189, #138 and #60

PR #286, opened 2023-05-29
Disable broken tests; force CI to run on every PR

cjbarth:
To improve code quality going forward, this PR will enable all tests to run on every PR. To get this to run the first time, broken tests have been disabled.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,19 +1,33 @@
-name: build
+name: Build Status
 on:
+  workflow_dispatch:
   push:
+    branches: [master]
   pull_request:
-    types: [opened, reopened]
+    branches: [master]
 
 jobs:
-  test:
+  build:
     runs-on: ubuntu-latest
+    env:
+      CI: true
+
     strategy:
       matrix:
-        version: [8, lts/*, latest]
+        node-version: [4, 6, 8, 10, 12, 14, 16, 18]
+
     steps:
-      - uses: actions/checkout@HEAD
-      - uses: actions/setup-node@HEAD
+      - uses: actions/checkout@v3
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v3
         with:
-          node-version: ${{ matrix.version }}
-      - run: npm install
+          node-version: ${{ matrix.node-version }}
+      - run: npm ci
       - run: npm test
+      - run: npm update
+      - run: npm ci
+      - run: npm test
+      - name: Codecov
+        uses: codecov/codecov-action@v3.1.1
+        with:
+          verbose: true
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "npm" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "monthly"
+    # Always increase the version requirement
+    # to match the new version.
+    versioning-strategy: increase
+  - package-ecosystem: "github-actions" # See documentation for possible values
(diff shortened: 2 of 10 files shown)


LoneRifle:
Could you clarify what you mean by broken tests? I just did `npm test` on my local dev and on a gitpod instance using Node 16 as well as Node 4, and all the tests pass.

cjbarth:
This is why we have code review :)

It turns out that `nodeunit` doesn't do globbing very well. Once I reverted the glob, all the tests started working. Sorry about that.

So, the only thing that this PR would do then is to make sure all tests run on every PR.

.github/workflows/ci.yml
```diff
@@ -1,19 +1,33 @@
...
+        uses: actions/setup-node@v3
         with:
-          node-version: ${{ matrix.version }}
-      - run: npm install
+          node-version: ${{ matrix.node-version }}
+      - run: npm ci
       - run: npm test
+      - run: npm update
+      - run: npm ci
+      - run: npm test
```
LoneRifle, line 29: Couple of things:
- Why the need to run `npm ci; npm test` twice? are we trying to test the codebase against the deps defined in package-lock as well as the deps after updating said lock?
- You may consider the use of `if` to [conditionally](https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution) run commands, so that for newer node versions we run `npm ci`, and for older ones we run `npm install`.
  - This is of course dependent on whether we want to continue supporting node 4 and 6. Limiting ourselves to node 8 would give us the freedom to rewrite the codebase with more modern language constructs (like async/await and `const` and `let` instead of `var`), but that depends on whether we get round to doing so.

.github/workflows/ci.yml
```diff
@@ -1,19 +1,33 @@
...
-  test:
+  build:
     runs-on: ubuntu-latest
+    env:
+      CI: true
+
     strategy:
       matrix:
-        version: [8, lts/*, latest]
+        node-version: [8.x, 10.x, 12.x, 14.x, 16.x, 18.x]
```
LoneRifle, line 17: - for conciseness, it should suffice to specify just `n` instead of `n.x`
- I would like to retain the `lts/*` and `latest` aliases, so that in the event that no one is available to update the list of versions, we have some assurance that we are testing against actively maintained versions of node

LoneRifle (approved):
Generally lgtm. Some concerns about the nature of the changes but non-blockers.

shunkica:
> It turns out that `nodeunit` doesn't do globbing very well. Once I reverted the glob, all the tests started working. Sorry about that.

Maybe this is an opportunity to move away from `nodeunit` as it is a dead project.

cjbarth:
@shunkica , I'd love to see a PR that does that. We currently use [Jasmine](https://jasmine.github.io/) in the other projects here, so it would be reasonable to move to that for consistency. However, anything would be better than `nodeunit` at this point.

.github/workflows/ci.yml
cjbarth: I'm acting as if we are going to do a `semver-major` release to drop older versions of Node. I really don't see a need to keep supporting something that old.


> are we trying to test the codebase against the deps defined in package-lock as well as the deps after updating said lock?

Yes, that is the idea. We had some problems with `node-saml` and `passport-saml` where people would report bugs that we couldn't reproduce. We found it was actually due to an incompatibility with a dependency. We actually found a problem with a TypeScript incompatibility/regression. Maybe I can mark these as `experimental` so that we don't fail the entire PR on them? I just don't want problem like that to remain hidden for very long, so I hesitate to do even that.

cjbarth:
@shunkica , would you be willing to put up a PR for a migration of the tests?

.github/workflows/ci.yml
cjbarth: For reference, I think that this might be good to keep in mind: https://github.com/actions/setup-node/issues/26#issuecomment-1148828915

cjbarth:
@LoneRifle , I think I've addressed all your concerns. Feel free to have another look at this. Your feedback is quite good. I look forward implementing some of this over in `node-saml` and `passport-saml`.

LoneRifle (approved):
lgtm, as discussed. Thanks for looking into this!

PR #287, opened 2023-05-29
Add support for linting

cjbarth:
This PR will add support for linting the code, but doesn't turn it on as part of the CI build. After this lands, another PR will land with all the code linted and the linting check turned on as part for the CI build.


.eslintignore
```diff
@@ -0,0 +1 @@
+node_modules
```
LoneRifle, line 1: Isn't node_modules [implicitly ignored](https://eslint.org/docs/latest/use/configure/ignore#the-eslintignore-file)?

.eslintrc.json
```diff
@@ -0,0 +1,18 @@
...
+    "node": true,
+    "mocha": true,
+    "es6": false
+  },
+  "root": true,
+  "plugins": ["prettier"],
+  "parserOptions": {
+    "ecmaVersion": 6
+  },
+  "extends": ["eslint:recommended", "prettier"],
```
LoneRifle, line 12: If you were to use say..
```suggestion
  "extends": ["eslint:recommended", "plugin:prettier/recommended"],
```

You might be able to do away with the prettier invocation after eslint within `npm run lint`. This would also help avoid any conflict between prettier and eslint by embedding prettier's rules/logic within eslint itself.

Not sure how this interacts with the prettier-plugin-packagejson though.

LoneRifle (approved):
lgtm otherwise

LoneRifle (approved):
lgtm

PR #289, opened 2023-05-31
Format code

cjbarth:
This is a collection of automatic code format changes from running `npm run prettier-format`. There are just a few changes in `example.js` and `README.md` that were manual (to disable rules and add code fences).

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -12,6 +12,7 @@
   "rules": {
     "no-console": "error",
     "no-unused-vars": "warn",
-    "no-prototype-builtins": "warn"
+    "no-prototype-builtins": "warn",
+    "one-var": ["warn", "never"]
   }
 }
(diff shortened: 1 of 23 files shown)


cjbarth:
I would eventually like to update the lint rules to be something like the following because the current code mixes many different styles:

```json
{
  "env": {
    "node": true,
    "mocha": true,
    "es6": false
  },
  "root": true,
  "parserOptions": {
    "ecmaVersion": 6
  },
  "extends": ["eslint:recommended", "prettier"],
  "rules": {
    "no-console": "error",
    "no-unused-vars": "warn",
    "no-prototype-builtins": "warn",
    "one-var": ["warn", "never"],
    "no-duplicate-imports": "error",
    "no-use-before-define": "error",
    "curly": "warn",
    "eqeqeq": ["error", "smart"],
    "no-var": "warn",
    "prefer-const":"warn"
  }
}
```

After this lands, I'll put up a PR for that so that we can properly discuss.

LoneRifle (approved):
lgtm; linting-related changes, I'm assuming no logic change. Relying on unit tests (which themselves have changed) to ensure correctness

Issue #77, opened 2015-10-04
fix travis builds


bjrmatos:
hi @yaronn, i noticed some time ago that the travis builds are broken because we are testing against an older version of node (0.6), can i update .travis.yml to test against 0.10 and 0.12?

yaronn:
good idea!

bjrmatos:
:+1:

PR #293, opened 2023-06-01
Make linting rules more strict

cjbarth:
This PR aims to improve code quality and consistency by tightening up some linting rules slightly. These rules require Node >=4 because of the rule to prefer `let` over `var`.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -12,6 +12,13 @@
   "rules": {
     "no-console": "error",
     "no-unused-vars": "warn",
-    "no-prototype-builtins": "warn"
+    "no-prototype-builtins": "warn",
+    "one-var": ["warn", "never"],
+    "no-duplicate-imports": "error",
+    "no-use-before-define": "error",
+    "curly": "warn",
+    "eqeqeq": ["error", "smart"],
+    "no-var": "warn",
+    "prefer-const":"warn"
   }
 }


cjbarth:
@yaronn , this is a draft right now because we don't pass all the rules. I'd like your feedback on two things:

1. Do these rules seem reasonable to you?
2. Do you think that setting the minimum require `engine` version in `package.json` to 4 would be a `semver-major` breaking change, or can we just do that right now. For referenced, our minimum required Node version is `>=0.4.0` which was [released](https://github.com/nodejs/node/releases/tag/v0.4.0) in 2011. It seems to me that perhaps when that was entered, the idea was to make the minimum required version `>=4.0.0` instead of `>=0.4.0`, the latter being a typo.

shunkica:
Maybe you have some other information but as far as I can tell @yaronn has not been active anywhere for over 5 years.

cjbarth:
@shunkica , you're right. I meant to do @LoneRifle  :)

LoneRifle:
Rules make sense to me. I think targeting 0.4.0 was what yaronn deliberately intended, see eg #77. That said, even with Travis, we've been testing against newer versions of Node, so perhaps it may make sense to tweak engine to line up with whatever we are supporting on our CI now, especially since the package's dependencies need at least Node 6

LoneRifle (approved):
lgtm

PR #294, opened 2023-06-05
Replace Nodeunit with Mocha

shunkica:


shunkica:
Hm... The tests are failing because Node.JS 12.17+ is required for Jasmine.

cjbarth:
I'm sorry @shunkica , I was working on another project that used Jasmine, this org uses Mocha. I've pushed a change to support Mocha. Thank you for your hard work!

cjbarth:
> Hm... The tests are failing because Node.JS 12.17+ is required for Jasmine.

@LoneRifle , this is going to be an ever bigger problem going forward. I suggest that we read [this](https://medium.com/the-node-js-collection/maintainers-should-consider-following-node-js-release-schedule-ab08ed4de71a) and decide what we'd like to do. I'm all for supporting `lts/-2`, but I don't like the idea of staying farther back than that. Thoughts?

LoneRifle:
@cjbarth I concur; I took the trouble to look at NodeJS' download [metrics](https://nodejs.org/metrics). It is worth noting that most downloads in the past two years or so are limited to `lts/-2`, thereabouts.

My only concern is within enterprise organisations, which tend to have very long change management cycles to bring in new software releases and hence a tendency to have very old releases of Node. But considering the number of security vulnerabilities found in older releases, perhaps these organisations would have moved quicker to get to newer ones.

cjbarth:
Reviewing this code is very difficult. I'm going to avoid the rename for the first pass, review it, land it, and then do a file rename.

cjbarth:
For reference, here is the log of tests from `nodeunit`. There are 232 assertions, so after the refactor, there should be 232 assertions, unless we have a clear reason why not.

```log
nodeunit ./test/canonicalization-unit-tests.js ./test/c14nWithComments-unit-tests.js ./test/signature-unit-tests.js ./test/saml-response-test.js ./test/signature-integration-tests.js ./test/document-test.js ./test/wsfed-metadata-test.js ./test/hmac-tests.js ./test/c14n-non-exclusive-unit-test.js


canonicalization-unit-tests.js
✔ Exclusive canonicalization works on xml with no namespaces
✔ Exclusive canonicalization works on inner xpath
✔ Exclusive canonicalization works on xml with prefixed namespaces defined in output nodes
✔ element used prefixed ns which is also the default
✔ Exclusive canonicalization works with default namespace for prefix
✔ Exclusive canonicalization works on xml with prefixed namespaces defined in output nodes. ns definition is not duplicated on each usage
✔ Exclusive canonicalization works on xml with prefixed namespaces defined in output nodes but before used
✔ Exclusive canonicalization works on xml with prefixed namespaces defined outside output nodes
✔ Exclusive canonicalization works on xml with prefixed namespace defined in inclusive list
✔ Exclusive canonicalization works on xml with multiple prefixed namespaces defined in inclusive list
✔ Exclusive canonicalization works on xml with prefixed namespace defined in inclusive list defined outside output nodes
✔ Exclusive canonicalization works on xml with prefixed namespace defined in inclusive list used on attribute
✔ Exclusive canonicalization works on xml with default namespace inside output nodes
✔ Exclusive canonicalization works on xml with multiple different default namespaces
✔ Exclusive canonicalization works on xml with multiple similar default namespaces
✔ Exclusive canonicalization works on xml with default namespace outside output nodes
✔ Exclusive canonicalization works when prefixed namespace is defined in output nodes not in the parent chain of who needs it
✔ Exclusive canonicalization works on xml with unordered attributes
✔ Exclusive canonicalization sorts upper case attributes before lower case
✔ Exclusive canonicalization removes Comments
✔ Exclusive canonicalization works on xml with attributes with different namespace than element
✔ Exclusive canonicalization works on xml with attribute values with special characters
✔ Exclusive canonicalization does not normalize whitespace characters into single spaces
✔ Exclusive canonicalization works on xml with element values with special characters
✔ Exclusive canonicalization preserves white space in values
✔ Exclusive canonicalization does not alter CR-NL (windows line separator) sequences
✔ Exclusive canonicalization preserves and encodes CR white space
✔ Exclusive canonicalization preserves white space between elements
✔ Exclusive canonicalization turns empty element to start-end tag pairs
✔ Exclusive canonicalization preserves empty start-end tag pairs
✔ Exclusive canonicalization with empty default namespace outside output nodes
✔ Exclusive canonicalization with empty default namespace inside output nodes
✔ The XML declaration and document type declaration (DTD) are removed
✔ Attribute value delimiters are set to quotation marks (double quotes)
✔ CDATA sections are replaced with their character content
✔ SignedInfo canonization
✔ Exclusive canonicalization works on complex xml
✔ Multiple Canonicalization with namespace definition outside of signed element
✔ Enveloped-signature canonicalization respects currentnode
✔ The XML canonicalization method processes a node-set by imposing the following additional document order rules on the namespace and attribute nodes of each element:   - An element's namespace and attribute nodes have a document order position greater than the element but less than any child node of the element.     Namespace nodes have a lesser document order position than attribute nodes.   - An element's namespace nodes are sorted lexicographically by local name (the default namespace node, if one exists, has no local name and is therefore lexicographically least).   - An element's attribute nodes are sorted lexicographically with namespace URI as the primary key and local name as the secondary key (an empty namespace URI is lexicographically least).     Lexicographic comparison, which orders strings from least to greatest alphabetically, is based on the UCS codepoint values, which is equivalent to lexicographic ordering based on UTF-8.
✔ saml attributed order (bug #25)
✔ Body Xml Element Canonicalization
✔ Overriding namespace in canonicalization

c14nWithComments-unit-tests.js
✔ Exclusive canonicalization works on xml with no namespaces
✔ Exclusive canonicalization works on inner xpath
✔ Exclusive canonicalization works on xml with prefixed namespaces defined in output nodes
✔ element used prefixed ns which is also the default
✔ Exclusive canonicalization works on xml with prefixed namespaces defined in output nodes. ns definition is not duplicated on each usage
✔ Exclusive canonicalization works on xml with prefixed namespaces defined in output nodes but before used
✔ Exclusive canonicalization works on xml with prefixed namespaces defined outside output nodes
✔ Exclusive canonicalization works on xml with prefixed namespace defined in inclusive list
✔ Exclusive canonicalization works on xml with multiple prefixed namespaces defined in inclusive list
✔ Exclusive canonicalization works on xml with prefixed namespace defined in inclusive list defined outside output nodes
✔ Exclusive canonicalization works on xml with prefixed namespace defined in inclusive list used on attribute
✔ Exclusive canonicalization works on xml with default namespace inside output nodes
✔ Exclusive canonicalization works on xml with multiple different default namespaces
✔ Exclusive canonicalization works on xml with multiple similar default namespaces
✔ Exclusive canonicalization works on xml with default namespace outside output nodes
✔ Exclusive canonicalization works when prefixed namespace is defined in output nodes not in the parent chain of who needs it
✔ Exclusive canonicalization works on xml with unordered attributes
✔ Exclusive canonicalization sorts upper case attributes before lower case
✔ C14N#WithComments retains Comments
✔ Exclusive canonicalization works on xml with attributes with different namespace than element
✔ Exclusive canonicalization works on xml with attribute and element values with special characters
✔ Exclusive canonicalization preserves white space in values
✔ Exclusive canonicalization preserves white space between elements
✔ Exclusive canonicalization turns empty element to start-end tag pairs
✔ Exclusive canonicalization preserves empty start-end tag pairs
✔ Exclusive canonicalization with empty default namespace outside output nodes
✔ Exclusive canonicalization with empty default namespace inside output nodes
✔ The XML declaration and document type declaration (DTD) are removed
✔ Attribute value delimiters are set to quotation marks (double quotes)
✔ CDATA sections are replaced with their character content
✔ SignedInfo canonization
✔ Exclusive canonicalization works on complex xml
✔ Multiple Canonicalization with namespace definition outside of signed element
✔ Enveloped-signature canonicalization respects currentnode
✔ The XML canonicalization method processes a node-set by imposing the following additional document order rules on the namespace and attribute nodes of each element:   - An element's namespace and attribute nodes have a document order position greater than the element but less than any child node of the element.     Namespace nodes have a lesser document order position than attribute nodes.   - An element's namespace nodes are sorted lexicographically by local name (the default namespace node, if one exists, has no local name and is therefore lexicographically least).   - An element's attribute nodes are sorted lexicographically with namespace URI as the primary key and local name as the secondary key (an empty namespace URI is lexicographically least).     Lexicographic comparison, which orders strings from least to greatest alphabetically, is based on the UCS codepoint values, which is equivalent to lexicographic ordering based on UTF-8.
✔ saml attributed order (bug #25)

signature-unit-tests.js
✔ signer adds increasing id attributes to elements
✔ signer adds references with namespaces
✔ signer does not duplicate existing id attributes
✔ signer adds custom attributes to the signature root node
✔ signer appends signature to the root node by default
✔ signer appends signature to a reference node
✔ signer prepends signature to a reference node
✔ signer inserts signature before a reference node
✔ signer inserts signature after a reference node
✔ signer creates signature with correct structure
✔ signer creates signature with correct structure (with prefix)
✔ signer creates correct signature values
✔ signer creates correct signature values using async callback
✔ correctly loads signature
✔ verify valid signature
✔ fail invalid signature
✔ allow empty reference uri when signing
✔ signer appends signature to a non-existing reference node
✔ signer adds existing prefixes
✔ creates InclusiveNamespaces element when inclusiveNamespacesPrefixList is set on Reference
✔ does not create InclusiveNamespaces element when inclusiveNamespacesPrefixList is not set on Reference
✔ creates InclusiveNamespaces element inside CanonicalizationMethod when inclusiveNamespacesPrefixList is set on SignedXml options
✔ does not create InclusiveNamespaces element inside CanonicalizationMethod when inclusiveNamespacesPrefixList is not set on SignedXml options
✔ adds attributes to KeyInfo element when attrs are present in keyInfoProvider

saml-response-test.js
✔ test validating SAML response
✔ test validating wrapped assertion signature
✔ test validating SAML response where a namespace is defined outside the signed element
✔ test reference id does not contain quotes
✔ test validating SAML response WithComments

signature-integration-tests.js
✔ verify signature
✔ verify signature of complex element
✔ empty URI reference should consider the whole document
✔ add canonicalization if output of transforms will be a node-set rather than an octet stream
✔ signature with inclusive namespaces
✔ signature with inclusive namespaces with unix line separators
✔ signature with inclusive namespaces with windows line separators
✔ should create single root xml document when signing inner node

document-test.js
✔ test with a document (using FileKeyInfo)
✔ test with a document (using StringKeyInfo)

wsfed-metadata-test.js
✔ test validating WS-Fed Metadata

hmac-tests.js
✔ test validating HMAC signature
✔ test HMAC signature with incorrect key
✔ test create and validate HMAC signature

c14n-non-exclusive-unit-test.js
✔ findAncestorNs: Correctly picks up root ancestor namespace
✔ findAncestorNs: Correctly picks up intermediate ancestor namespace
✔ findAncestorNs: Correctly picks up multiple ancestor namespaces declared in the one same element
✔ findAncestorNs: Correctly picks up multiple ancestor namespaces scattered among depth
✔ findAncestorNs: Correctly picks up multiple ancestor namespaces without duplicate
✔ findAncestorNs: Correctly eliminates duplicate prefix
✔ findAncestorNs: Exclude namespace which is already declared with same prefix on target node
✔ findAncestorNs: Ignores namespace declared in the target xpath node
✔ C14n: Correctly picks up root ancestor namespace
✔ C14n: Correctly picks up intermediate ancestor namespace
✔ C14n: Correctly picks up multiple ancestor namespaces declared in the one same element
✔ C14n: Correctly picks up multiple ancestor namespaces scattered among depth
✔ C14n: Correctly picks up multiple ancestor namespaces without duplicate
✔ C14n: Correctly eliminates duplicate prefix
✔ C14n: Exclude namespace which is already declared with same prefix on target node
✔ C14n: Preserve namespace declared in the target xpath node
✔ C14n: Don't redeclare an attribute's namespace prefix if already in scope
✔ C14n: Don't declare an attribute's namespace prefix if in scope from parent

OK: 232 assertions (1031ms)
```

shunkica:
[You would need something like this to count the number of assertions made](https://www.chaijs.com/plugins/chai-assertions-count/) I have not tested it.

cjbarth (dismissed):
All the tests look like they were ported without any changes.

PR #298, opened 2023-06-06
Adjust references for `node-saml` organization

cjbarth:
The README badge was pointing to the wrong location and the NPM package name was scoped, which it doesn't seem like we're using yet. We should discuss if we want to move this NPM package to be a scopped one and deprecate the unscopped version. For now, we have to unscope it so that we can do a release.


cjbarth:
@LoneRifle @markstos , I'd like to hear your thoughts on scopped vs. unscopped for this package.

PR #299, opened 2023-06-06
Target an LTS version of Node

cjbarth:
This is a `semver-major` change that will bring the dependencies up to supported versions and more easily allow migration of tests to supported test frameworks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "test": "nodeunit ./test/canonicalization-unit-tests.js ./test/c14nWithComments-unit-tests.js ./test/signature-unit-tests.js ./test/saml-response-test.js ./test/signature-integration-tests.js ./test/document-test.js ./test/wsfed-metadata-test.js ./test/hmac-tests.js ./test/c14n-non-exclusive-unit-test.js"
   },
   "dependencies": {
-    "@xmldom/xmldom": "0.8.7",
+    "@xmldom/xmldom": "^0.8.8",
     "xpath": "0.0.32"
   },
   "devDependencies": {
@@ -50,6 +50,6 @@
     "release-it": "^15.11.0"
   },
   "engines": {
-    "node": ">=4.0.0"
+    "node": ">=14"
   }
 }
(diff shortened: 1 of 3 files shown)


cjbarth:
This will use `lts/*`, `lts/-1` and `lts/-2`, but will do so via `[14, 16, 18]`, which allows us to update `package.json`'s `engine` as we move forward. This also tests `latest`, so we make sure that we see any coming issues as `node` is upgraded.

parallels999:
Hi, when will there be a new release with these changes?

cjbarth:
As soon as we finish the TS conversion... which is dependent on https://github.com/goto100/xpath/pull/116 landing. We're getting close. FWIW, this release is holding up the v5 release of `passport-saml` and `node-saml`, so there is strong motivation to get this along.

parallels999:
@cjbarth so, is possible https://github.com/node-saml/xml-crypto/pull/333 Bump @xmldom/xmldom to 0.8.8 for **V3**?

cjbarth:
Sure, just make a PR against `master`, we'll land it, and then make the same thing against `3.x`.

parallels999:
that's the point, master already has this change, **ON THIS  PR**
https://github.com/node-saml/xml-crypto/blob/131bcf97259a4023323cc93ebc5aa88290716832/package.json#L38

package.json
```diff
@@ -35,7 +35,7 @@
     "test": "nodeunit ./test/canonicalization-unit-tests.js ./test/c14nWithComments-unit-tests.js ./test/signature-unit-tests.js ./test/saml-response-test.js ./test/signature-integration-tests.js ./test/document-test.js ./test/wsfed-metadata-test.js ./test/hmac-tests.js ./test/c14n-non-exclusive-unit-test.js"
   },
   "dependencies": {
-    "@xmldom/xmldom": "0.8.7",
+    "@xmldom/xmldom": "^0.8.8",
```
parallels999, line 38: @cjbarth look here

PR #283, opened 2023-05-24
Add support for appending attributes to KeyInfo element

shunkica:
Added a new optional property, attrs, to the KeyInfo class.
The attrs property is an object that holds key-value pairs representing the attributes and their corresponding values. 
When present, these attributes will be appended to the KeyInfo element during XML generation.

Example:

```
sig.keyInfoProvider = {
    getKeyInfo: () => {
        return `<wsse:SecurityTokenReference xmlns:wsse="http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-secext-1.0.xsd" xmlns:wsu="http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-utility-1.0.xsd" wsu:Id="STR-${securityTokenReferenceGuid}">`
            + `<wsse:Reference URI="#X509-${randomUUID()}" ValueType="http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-x509-token-profile-1.0#X509v3"/>`
            + `</wsse:SecurityTokenReference>`
    },
    getKey: () => Buffer.from(privateKey),
    attrs: {
        Id: `KI-${randomUUID()}`,
    }
};
```

Result:

```
<ds:KeyInfo Id="KI-a8bb2a1a-32dd-449e-9bf0-187dd8ea4ee6">
  <wsse:SecurityTokenReference wsu:Id="STR-0096ce19-3fbf-414f-94e9-f7a0d601b3f4">
    <wsse:Reference URI="#X509-8cc9b13b-ee21-426e-b8d1-75cddddbe6e5"
      ValueType="http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-x509-token-profile-1.0#X509v3"/>
  </wsse:SecurityTokenReference>
</ds:KeyInfo>
```

Issue #240, opened 2021-12-30
Suggestion: a developer friendly, secure API


lukaszmakuch:
Hi! 👋

I'd like to thank everyone who has ever worked on this library, as it has helped me build things I wouldn't have built otherwise. 

Because I like it so much, I must tell you that I noticed some security issues which I believe could be avoided if the API was simpler.

**I'd like to propose adding a high-level API that makes our apps harder to hack.** Here's how it'd look:
```js
node = getTheOnlySignedNodeOrNull(xml, publicKey)
```
and
```js
nodesArray = getSignedNodes(xml, publicKey)
```
You can read the rationale behind it in this blog post - [Your XML security library is sabotaging your work. Here's what you can do about it](https://coder.earth/post/your-xml-lib-is-sabotaging-your-work). It's evident verifying signatures poses a real challenge. An example is described in a post titled [Are XML Signatures secure?](https://coder.earth/post/are-xml-signatures-secure).

The current, low-level API doesn't have to be removed, so everything can be backward-compatible.

The new, safer API can be built on top of the existing one.

Should we help developers deliver secure solutions?

djaqua:
Yes, I love this, personally, but I think it's worth having a discussion around making the high level API as flexible as possible.  We're in the process of knocking out some real low-level pain points and potentially some redesign. 

The API you're proposing is basically a facade over the package that ensures that clients successfully accomplish what they mean to, right?

lukaszmakuch:
You're right @djaqua. What I'm proposing is a facade over the current API. I have actually implemented something like that for a client of mine so that reading "safe" data from XMLs is not a footgun.

I'd love to hear more about the ideas for the potential redesign. What are the goals you're aiming for with this redesign?

cjbarth:
@djaqua @lukaszmakuch , how's this discussion going?

lukaszmakuch:
> @djaqua @lukaszmakuch , how's this discussion going?

I don’t have any updates.

cjbarth:
Would either @djaqua or @lukaszmakuch like to propose such an API, either here or in a PR so that this can keep moving forward?

djaqua:
@cjbarth & @lukaszmakuch -- check out #304 
I think the Facade would easily and effectively knock out high level goals 1 & 2. I would be glad to donate the Facade I have made for my own use. It would take me an evening or two to clean it up and add JSDocs. But .... I don't want to start coding until we have a real serious discussion about where we want to take this project.

[...]

PR #301, opened 2023-06-08
Add support for <X509Certificate /> in <KeyInfo />; remove `KeyInfoProvider`

cjbarth:
This [PR](https://github.com/node-saml/node-saml/pull/36) makes reference to how our handling of certificates in `<KeyInfo />` is incorrect. The documentation for this is [here](https://www.w3.org/TR/2008/REC-xmldsig-core-20080610/#sec-X509Data).

Further information about this can be found in [this](https://github.com/node-saml/xml-crypto/pull/295#discussion_r1220100389) comment thread.


LoneRifle (changes requested):
lgtm otherwise

index.d.ts
```diff
@@ -291,102 +291,52 @@ export interface KeyInfoProvider {
...
    * @return An XML string representation of the contents of a KeyInfo element.
    */
-  getKeyInfo(key?: string, prefix?: string): string;
+  makeKeyInfo(cert?: string | string[], prefix?: string): string;
 
   /**
    * An optional dictionary of attributes which will be added to the KeyInfo element.
    */
-  attrs?: { [key: string]: string };
+  attrs?: { [cert: string]: string };
```
shunkica, line 308: Revert this back to key

shunkica, line 308:
```suggestion
  attrs?: { [key: string]: string };
```

lib/file-key-info.js
```diff
@@ -1,17 +1,16 @@
...
+ * A key info provider implementation for certificates in files
  *
  * @param {string} file path to public certificate
  */
-function FileKeyInfo(file) {
-  var key = fs.readFileSync(file);
-  StringKeyInfo.apply(this, [key]);
+class FileKeyInfo extends KeyInfo {
+  constructor(file) {
+    const cert = fs.readFileSync(file);
```
shunkica, line 11: The content of the X509Certificate element shouldn't have the `-----BEGIN CERTIFICATE-----` and `-----END CERTIFICATE-----` tags.
So it should either be stated that you expect a base64 encoded DER as input, or remove the tags if this is a PEM certificate.
Also, if it is a PEM file, it can contain multiple certificates, so you should parse them and put them in an array.

lib/key-info.js
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * A basic, string-based, implementation of a KeyInfoProvider
+ *
+ * @param {string} cert the public certificate
+ * @implements {KeyInfoProvider}
+ */
+class KeyInfo {
+  constructor(cert) {
+    this.cert = cert;
```
shunkica, line 11: Specify this should be base64 DER, and not PEM, or strip the tags from PEM.

shunkica (changes requested):

lib/key-info.js
```diff
@@ -0,0 +1,55 @@
...
+   * Builds the contents of a KeyInfo element as an XML string.
+   *
+   * For example, if the value of the prefix argument is 'foo', then
+   * the resultant XML string will be "<foo:X509Data></foo:X509Data>"
+   *
+   * @param cert the public key as a string or array of strings (see https://www.w3.org/TR/2008/REC-xmldsig-core-20080610/#sec-X509Data)
+   * @param prefix an optional namespace alias to be used for the generated XML
+   * @return an XML string representation of the contents of a KeyInfo element
+   */
+  makeKeyInfo(cert = "", prefix = "") {
```
shunkica, line 23: I am not a fan of this rename. If your concern is the overlap with `SignedXml.getKeyInfo` method, maybe you should consider renaming the other one.

lib/key-info.js
```diff
@@ -0,0 +1,55 @@
...
+    prefix = prefix ? prefix + ":" : prefix;
+
+    if (typeof cert !== Array) {
+      cert = [cert];
+    }
+
+    const x509Certs = cert.map((cert) => `<X509Certificate>${cert}</X509Certificate>`);
```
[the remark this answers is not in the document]
shunkica: Just join with empty string instead of newline. I don't think this library puts newlines anywhere so why start now.

lib/signed-xml.js
```diff
@@ -377,8 +383,8 @@ SignedXml.prototype.checkSignature = function (xml, callback) {
     }
   }
 
-  this.signingKey = this.keyInfoProvider.getKey(this.keyInfo);
-  if (!this.signingKey) {
+  this.signingCert = this.keyInfoProvider.getCert(this.keyInfo);
```
shunkica, line 386: This is confusing.
It should be `this.signingCert = this.keyInfoProvider.getCert(this.keyInfo);`
And then in `validateSignatureValue` also change to `this.signingCert`:

Overall more clear lines of separation are needed in the library between what the private signing key is and what the public certificate is.

shunkica (changes requested):

index.d.ts
```diff
@@ -291,102 +291,52 @@ export interface KeyInfoProvider {
...
 }
 
-/**
- * The FileKeyInfo class loads the certificate from the file provided in the constructor.
- */
-export class FileKeyInfo implements KeyInfoProvider {
+export class KeyInfo implements KeyInfoProvider {
   /**
-   * The path to the file from which the certificate is to be read.
+   * The certificate  to be used for signing.
```
shunkica, line 313: Misleading. The private key is used for signing, the certificate is used for validating the signature.

index.d.ts
cjbarth: But this isn't a key to be provided here, it is a cert as it is part of the `<X509Certificate />` payload. Am I missing something?

lib/file-key-info.js
cjbarth: Do you think a comment to that effect would be sufficient at this time?

lib/key-info.js
cjbarth: Do you think a comment to that effect would be sufficient at this time?

lib/signed-xml.js
cjbarth: I don't follow your concern. It seems that it is already as you suggest. On line 478, we pass `this.signingCert || this.signingKey` because the downstream functions will accept a public key or a private key, because a public key can be derived from a private key. Since we aren't working with "public keys" directly in this codebase, we're sending `signingCert` instead, since a cert only contains the public part.

index.d.ts
cjbarth: You're statement is true. I will review the code comments.

index.d.ts
shunkica: The `key ` part of the index signature `[key: string]` is just a placeholder value, typically people just put `key`,
If you want to be more precise you could put something like `attrName` there.
And the values of attr are not part of the X509Certificate payload, they are attributes of the KeyInfo element.
For example if  `attr = {Id:"12345", foo:"bar"}` then `<KeyInfo Id="12345" foo="bar">`

cjbarth: Right... a TypeScript thing... that is a crazy oversight on my part. Too much language hopping lately :)

lib/signed-xml.js
shunkica: We already have these two properties in SignedXml. We have `this.signingKey`, which is meant to contain the private key, and we have `this.signingCert`, which contains the certificate (public key).
So why put the public key you get from `keyInfoProvider.getCert(..)` into the property which is meant to contain the private key?

lib/key-info.js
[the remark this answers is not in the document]
cjbarth: I agree.

lib/signed-xml.js
cjbarth: I don't. The line as it current is reads: `this.signingCert = this.keyInfoProvider.getCert(this.keyInfo);` The previous name was very confusing on this matter, thus my correcting it. Can you clarify further what you're talking about? Maybe there is a line that I'm misreading.

lib/file-key-info.js
shunkica: > Do you think a comment to that effect would be sufficient at this time?

Honestly, no. You should really modify the function to read a PEM chain from a file.
There is no point in this class otherwise.

The existing FileKeyInfo was meant to work with PEM files as well. From the docs:  

> A basic key info provider implementation using fs.readFileSync(file), is constructed using new FileKeyInfo([file]) where:
>     file - a path to a pem encoded certificate

Most people are used to working with PEM files. Nobody has a base64 encoded DER certificate file, and even if you make it, you can only put 1 certificate in there.
You added the functionality for the certificate to be an array of certificates, so this would be in line with that change.

I suggest you take a PEM file as input, and parse the certificate(s) inside into an array.

shunkica (changes requested):

lib/file-key-info.js
```diff
@@ -1,17 +1,16 @@
...
  *
- * @param {string} file path to public certificate
+ * @param {string} file path to public certificate, in base64 DER format, not PEM format
  */
-function FileKeyInfo(file) {
-  var key = fs.readFileSync(file);
-  StringKeyInfo.apply(this, [key]);
+class FileKeyInfo extends KeyInfo {
+  constructor(file) {
+    const cert = fs.readFileSync(file);
```
shunkica, line 11:
```suggestion
      var pem = fs.readFileSync(file, 'utf8');
      var certificates = pem.match(/-----BEGIN CERTIFICATE-----[^-]*-----END CERTIFICATE-----/g);
      super(certificates.map(function(certificate) {
        return certificate.replace(/-----BEGIN CERTIFICATE-----|-----END CERTIFICATE-----|\n|\r/g, '');
      }));
```

cjbarth: That breaks tests since some of the certificates are stored in a binary format, so I didn't want to have to change tests or break compatibility with that format of cert.

shunkica: I noticed there are no tests for the KeyInfo class.
Can you make a test where it accepts string/array of string and then sign/validate?

cjbarth: We shouldn't have to test it separately as long as we have complete coverage. The code coverage scanning I've added helps us to see that we're pretty covered here. There are lots of tests that do just what you say but tests more parts of the code than just `KeyInfo`. Is there any part of the coverage that you feel is missing?

Honestly, we could probably check to make sure that the `<X509Certificate />` element is there. I don't know if I have time to write that test anytime soon. I've spent more on this than I wanted to. Do you have time to contribute a test @shunkica ?

lib/key-info.js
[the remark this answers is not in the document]
djaqua: Sharp catch, @LoneRifle :ok_hand:

lib/file-key-info.js
cjbarth: Please have a look at the changes I've made and see if they address your concerns.

lib/file-key-info.js
shunkica: Let's say I want to add the entire certificate chain during the signature process to the KeyInfo element using one of the providers you created.

I have my certificate chain in pem format, and my private key in pem format.

```
    var xml = "<root><x /></root>";
    var sig = new SignedXml();
    sig.signingKey = fs.readFileSync("./test/static/myprivate.pem");
    sig.keyInfoProvider = new FileKeyInfo("./test/static/mypublic_chain.pem");
    sig.computeSignature(xml);
    var signedXml = sig.getSignedXml();
```
This results in:
```
TypeError: Cannot read properties of null (reading 'match')
    at FileKeyInfo.getKeyInfoContent (lib\key-info.js:65:17)
    at SignedXml.getKeyInfo (lib\signed-xml.js:908:33)
    at SignedXml.computeSignature (lib\signed-xml.js:815:24)
```

shunkica: Here are some of my opinions regarding what the the implementation of the default KeyInfo classes should be.

**During signature:**
- the user can load certs in **PEM string** into the `KeyInfo | FileKeyInfo` constructor (it can be a single cert or a chain)
> from crypto.createPrivateKey(key) -> If key is a string or Buffer, format is assumed to be 'pem'
- only `getKeyInfoContent()` used during xml generation - how do we get the certs inside getKeyInfoContent?
  - we already have `this.certs`, so we can just load them from there
  - the cert parameter is not needed in this implementation and whatever is passed to it should be ignored by it

**During validation:**

- only `getCert()` is used.  It must return a PEM string or buffer

> from crypto.createPublicKey(key) -> Creates and returns a new key object containing a public key. If key is a string or Buffer, format is assumed to be 'pem'

- keyInfo is used to get the public certificate from the keyInfo, but it can also be passed a custom public certificate 
  - if user created the instance with `new KeyInfo()` or `new KeyInfo(null)` - > `getCert(keyInfo)` extracts the cert from the first X509Certificate element
  - if user created the instance with `new KeyInfo(cert)`
      - > we should assume the user wants to validate the signature against the specific public certificate they passed to the constructor,
	  so getCert should return this.cert[0], regardless of if it was passed a parameter or not.
	  
The problem is that in this case, the HMAC signature will not work, because it doesn't use PEM keys, and that is why the tests were failing.
So what is the solution to that? Probably drop support for HMAC, or have a completely different workflow for using it.
HMAC is a symmetric mechanism where the same key is used to generate and verify the signature,
and thus doesn't fit in the standard public key infrastructure that XMLDSIG is designed around.

cjbarth: Honestly, I thought about this a lot too. It seems to me that this entire thing should be static. If they call `getKeyInfoContents()`, they can pass in the certs they want. If the call `getCert()`, they can optionally specify a fallback public or private key (since a public key can be derived from a private key) that will be used if one can't be extracted from the `KeyInfo` block provided. That would solve lots of our problems and it would push the entire concept of `FileKeyInfo` out the window as file handling would be the consumer's problem (as it probably should be).

Thoughts?

shunkica: > Honestly, I thought about this a lot too. It seems to me that this entire thing should be static. If they call `getKeyInfoContents()`, they can pass in the certs they want. If the call `getCert()`, they can optionally specify a fallback public or private key (since a public key can be derived from a private key) that will be used if one can't be extracted from the `KeyInfo` block provided. That would solve lots of our problems and it would push the entire concept of `FileKeyInfo` out the window as file handling would be the consumer's problem (as it probably should be).
> 
> Thoughts?

I'm tapped out for thoughts currently, I don't think I have anything more to add than what I wrote above.
In any case, the HMAC thing needs dealing with as outlined in my comment. Those should not be used with KeyInfoProvider.

cjbarth: @LoneRifle , perhaps you have fresh thoughts. We've known this part of the code is broken for a long time and it would be nice to fix it correctly now.

LoneRifle: A lot has been discussed though. I've never understood why a class was ever needed to hold key info. It's not like the class state (ie, the cert) is ever manipulated, or the helper methods produce an output that meaningfully combines class state with inputs.

I'm likewise inclined to break the class and turn it into a helper module, or if we can afford to, collapse its methods into SignedXml. We then manipulate certs solely in-memory, and leave it to the caller to figure out how to provide them, ie as-is, reading a file, or even async I/O.

shunkica: I'm guessing it was made that way so it is easy for the user to make their own implementation of the key provider.

djaqua:
@cjbarth @LoneRifle @shunkica 

Okay, again, sorry for the late response here, took me a minute to catch up with what's going on.

Here's my high level thoughts (without writing a novel like I did for that other conversation):

1. I love the `getKeyInfoContent` function
2. I think it offering X509Certificate functionality out of the box is a **great** idea
3. I think it binding KeyInfo to the X509Certificate is a road we should not go down

I know you guys have lives and that the last thing any developer wants is another meeting, but I think it would be really really beneficial if we have a discussion about this project and figure out how to tactfully cross some bridges. 

If you guys don't want a meeting, I'll have to write a novel to explain why I think we should not bind KeyInfo to X509Certificate 

But I'm not a complainer; I wouldn't drag my feet on this if I didn't have a reasonable solution. The gist of my solution is that we could and should knock off the two suggestions that I wrote in my novel yesterday before proceeding with anything X509Certificate:

1. Deprecate SignedXml.signingKey in favor of using SignedXml.privateKey and SignedXml.publicKey
2. Implement a Facade like [lukaszmakuch](https://github.com/lukaszmakuch) suggested in [this discussion](#240 )

So then we could move onto X509Certificate by replacing KeyInfoProvider altogether. The totally bodacious paradigm would be to have something called `X509CertificateKeyInfoBuilder` with the function `getKeyInfoContent` implemented as you have it.

LoneRifle:
> I'll have to write a novel to explain why I think we should not bind KeyInfo to X509Certificate

@djaqua might actually be helpful, so we can read async and come to any proposed meeting with working knowledge.  It also documents in some manner a critical decision we are about to make with this package.

LoneRifle:
I think it is very clear that we have issues to address with respect to the design of this package (specifically around KeyInfo generation) before we can consider implementing support for X509 certificates.

@cjbarth - I would like to suggest that maybe we close this PR for now, and overhaul KeyInfoProvider/SignedXml. Once that is in place, we might be in a better position to add X509 support.

wdyt?

lib/file-key-info.js
LoneRifle: While true, thus far, the only practical example I've seen is from your PR comments in #283.

Maybe what might be best is a property specified by SignatureAlgorithm implementations that states whether an algorithm employs either public-key cryptography or a secret. SignatureAlgorithms that state they use shared secrets would then go on a separate code path from those that don't, as you outlined above.

Thoughts?

shunkica: > While true, thus far, the only practical example I've seen is from your PR comments in https://github.com/node-saml/xml-crypto/pull/283.

I'm not sure I follow what you are saying. I think in most cases users are using a custom `keyInfoProvider`. It's also documented [here](https://github.com/node-saml/xml-crypto#customizing-algorithms)

> Maybe what might be best is a property specified by SignatureAlgorithm implementations that states whether an algorithm employs either public-key cryptography or a secret. SignatureAlgorithms that state they use shared secrets would then go on a separate code path from those that don't, as you outlined above.
> 
> The package user will also be instructed to specify their secret explicitly as the private key in SignedXml, to drive home the point that the secret is meant to be kept as one. For schemes that employ public-key cryptography, both private and public keys will be specified.

That is a possible solution, but I think this conversation is better suited for issue #303
I don't see it as directly related to this PR.

LoneRifle: Nothing further to add at this point =)

cjbarth: > I don't see any tests where KeyInfo or FileKeyInfo is used during the signature process. There are only tests using FileKeyInfo during validation process.

I've corrected your test and adjusted the code for `KeyInfo`. It now works and we have some coverage. Thank you for that contribution.

cjbarth:
> 1. Deprecate SignedXml.signingKey in favor of using SignedXml.privateKey and SignedXml.publicKey

This seems reasonable. I've pushed up a change to make this `privateKey` and `publicCert` instead of `signingKey` and `signingCert`, respectively.

cjbarth:
> I would like to suggest that maybe we close this PR for now, and overhaul KeyInfoProvider/SignedXml. Once that is in place, we might be in a better position to add X509 support.

I'd rather see if we can find some middle ground with this PR. I haven't seen any concrete proposals that we can action on yet. This is the closest thing we have to something that works. The HMAC thing is the biggest outlier, and I haven't seen consensus that we want to drop support for that.

cjbarth:
I think the next step is to make it so this class can't be instantiated and adjust the tests to match that. Feel free to push up a commit to this branch that does that. Once we do that, the concept of `FileKeyInfo` will be meaningless and can be removed. Users will be responsible for reading their own files. We already have helper methods to convert between PEM and DER, so that can help users.

I suggest starting with something like this and going from there:

```javascript
class KeyInfo {
  constructor() {
    throw Error("This class is static; you may not instantiate it.");
  }
...
```

If we leave it as a class, then it still allows people to `extends` it should they need to which allows a 'plugable' feel to this part of the library.

cjbarth:
The other option that was presented here, which I'm starting to like the more that I think about it, is to put most of the new stuff in `KeyInfo` into a utility class and then add two properties to `SignXml`: `getKeyInfoContent` and `getCertFromKeyInfo`. We can provide defaults for these properties, but still allow others to override them. They they can decide what they want their own cert to do when validating. Thoughts?

lib/file-key-info.js
shunkica: Can you explain why the public cert needs to be set twice?

cjbarth: The public cert needs to be set twice, because we don't use one of them. The public cert set in the ctor doesn't do anything when building the `<X509Certificate />` element, the only thing that does anything is the `publicCert` property. That is one of the reasons that I suggest that we just get rid of the ctor and make this entirely static or just remove the entire class and put two properties on `SignXml` for the two functions we need, with defaults, so they can override as needed.

cjbarth:
I can revert this last commit if you all don't like the direction, but I think this really simplifies things. I have a few tests to review after this bulk refactor; I'm not sure why they are failing. Let me know what you think of this direction.

djaqua:
> 

@cjbarth & @LoneRifle 

The more I thought about it, the more I realized that I don't really need to write a novel to explain, so let me attempt to do so concisely:

X509 is one of potentially many paradigms that XML digital signatures can work with. Another one right off the top of my head is the `SecurityTokenReference` element (a child of `KeyInfo`) in conjunction with a `BinarySecurityToken` from the WS-Security specs.

I would not default to using X509 because the default behavior of the current implementation is to skip building a KeyInfo element altogether (`SignedXml.prototype.getKeyInfo`). If we default to building an X509 thingamajig, that could purposelessly screw up clients who may have written tests to validate their XML against their own schemas. I do understand that this is an inherently breaking change so their stuff is going to break anyways, but we should give clients the option to skip building a KeyInfo element since its not strictly required by the schema.

As a reasonable best-of-all-worlds solution, I would like to propose the following (and see the big NOTE below):
- define a KeyInfoStrategy class that does nothing by default, and default to using this when a client doesn't specify otherwise
- define an X509KeyInfoStrategy class that builds the X509 stuff; make clients manually choose this 

My main hesitation to get onboard with these high quality changes (they really are -- I'm not here to knock the code or the idea at all) is that, learning from our experience with `KeyInfoProvider`, we would do well to be very annoyingly thoughtful about the consequences of our decisions around the complexity-inducing `KeyInfo` element. Some things to be annoyingly thoughtful about:

- method signatures (i.e. function parameter lists) -- these should be implementation-agnostic
- method parameter and return types -- all strings? all xpath objects (e.g. Node, Element, etc)? combinations thereof? 
- method names -- I love `getCertFromKeyInfo` but I would prefer `createKeyInfoContent` to `getKeyInfoContent` (see the NOTE below)

That's pretty much my thoughts, here. I'm neither a dictator nor in charge of this project, so if there is disagreement about these concerns, lets continue the discussion :smiley: 

NOTE
I am very strongly opinionated about naming conventions. I welcome discussions and I'm open to alternative suggestions, so if anyone else feels strongly about such things, lets have the discussions :smile: , otherwise, if it's all the same to y'all, lets just do it my way :smiling_imp: :laughing:

djaqua:
And yes I am aware that I said "concisely" and proceeded to write a novel anyway :sweat_smile:

lib/file-key-info.js
LoneRifle: This overlooks the fact that there are other kinds of KeyInfo however, as @djaqua points out, and as @shunkica has shown in an example implementation of KeyInfoProvider. Extending SignXml to change the key provider might not be a good idea given that signing and verifying otherwise remains largely the same. That said, I'm okay for this to be done in the interim, while we figure out how to rebuild the KeyInfoProvider and/or SignXml's use of it.

After that, in the long run, SignXml can delegate to KeyInfoProvider or its successor once again.

cjbarth:
I've fixed the tests and taken your naming suggestion @djaqua.

I'd like to hear more about what you have to say about method signatures. I'm inclined to pass an object with properties to those functions, thereby allowing a custom implementation to work with other values as needed. As the need arises, we can easily, and without breaking, add more properties to the object we send for use-cases we don't see yet.

cjbarth:
@djaqua , have a look at that latest commit and let me know if that will address your concerns.

cjbarth:
> I am very strongly opinionated about naming conventions. I welcome discussions and I'm open to alternative suggestions, so if anyone else feels strongly about such things, lets have the discussions 😄 , otherwise, if it's all the same to y'all, lets just do it my way 😈 😆

I was thinking more about this, and it would be nice to have a convention. I did some googling and came across [this](https://github.com/kettanaito/naming-cheatsheet), which seems like a well-liked (12.7k stars) opinion. @djaqua , do you have a different reference you use? If not, I propose to avoid bike-shedding and use this.

README.md
```diff
@@ -254,23 +253,9 @@ var SignedXml = require("xml-crypto").SignedXml,
...
-    prefix = prefix ? prefix + ":" : prefix;
-    return "<" + prefix + "X509Data></" + prefix + "X509Data>";
-  };
-  this.getKey = function (keyInfo) {
-    //you can use the keyInfo parameter to extract the key in any way you want
-    return fs.readFileSync("key.pem");
-  };
-}
-```
+Implement it if you want to create a signature with a KeyInfo section, or you want to read your key in a different way then the built-in file-read or string methods. See the implementation in `string-key-info.js` for more information.
```
shunkica, line 256: The documentation on customizing the content of the KeyInfo element was removed.
There is only a reference to the `string-key-info.js` file which no longer exists.

lib/signed-xml.js
```diff
@@ -316,6 +319,9 @@ function SignedXml(idMode, options) {
   this.idAttributes = ["Id", "ID", "id"];
   if (this.options.idAttribute) this.idAttributes.splice(0, 0, this.options.idAttribute);
   this.implicitTransforms = this.options.implicitTransforms || [];
+  this.composeKeyInfoContent = SignedXml.composeKeyInfoContent;
+  this.getCertFromKeyInfo = SignedXml.getCertFromKeyInfo;
+  this.attrs = {};
```
shunkica, line 324: Bad name. Suggest `keyInfoAttrs` or similar

shunkica:
Reviewing the code is quite time-consuming due to the extensive changes it incorporates simultaneously, many of which are not related to the original intention of the PR of generating the X509Certificate element within the KeyInfo. (eg: renames and removal of the keyInfoProvider). Going forward, I recommend considering a more incremental approach, implementing smaller changes instead of a large number of modifications all at once.

cjbarth:
> Reviewing the code is quite time-consuming due to the extensive changes it incorporates simultaneously, many of which are not related to the original intention of the PR of generating the X509Certificate element within the KeyInfo. (eg: renames and removal of the keyInfoProvider). Going forward, I recommend considering a more incremental approach, implementing smaller changes instead of a large number of modifications all at once.

I'll gladly break it up. Let me just to the `KeyInfoProvider` stuff here. Sound good?

index.d.ts
```diff
@@ -129,6 +129,16 @@ export interface TransformAlgorithm {
...
  */
+
+/**
+ * @param cert the certificate as a string or array of strings (see https://www.w3.org/TR/2008/REC-xmldsig-core-20080610/#sec-X509Data)
+ * @param prefix an optional namespace alias to be used for the generated XML
+ */
+export interface ComposeKeyInfoContentArgs {
+  cert: string;
+  prefix: string;
+}
```
shunkica, line 140: The documentation states string or array of strings, but the parameter only accepts string.

lib/signed-xml.js
```diff
@@ -316,6 +313,9 @@ function SignedXml(idMode, options) {
   this.idAttributes = ["Id", "ID", "id"];
   if (this.options.idAttribute) this.idAttributes.splice(0, 0, this.options.idAttribute);
   this.implicitTransforms = this.options.implicitTransforms || [];
+  this.composeKeyInfoContent = SignedXml.composeKeyInfoContent;
```
shunkica, line 316: I'd like to share a perspective from a non-native English speakers viewpoint on this decision. The term "compose" seems slightly more unfamiliar to me. I believe the earlier choice of `getKeyInfoContent` is somewhat more conventional and concise.

lib/signed-xml.js
```diff
@@ -367,19 +418,13 @@ SignedXml.prototype.checkSignature = function (xml, callback) {
...
-  this.signingKey = this.keyInfoProvider.getKey(this.keyInfo);
-  if (!this.signingKey) {
-    var err2 = new Error("key info provider could not resolve key info " + this.keyInfo);
+  const signingCert = this.getCertFromKeyInfo(this.keyInfo) || this.signingCert;
+  if (!signingCert) {
+    var err2 = new Error(
+      "key info provider could not resolve key info " +
+        this.keyInfo +
+        "\n and no publicCert provided"
+    );
```
shunkica, line 427: This should probably not throw an error if `signingKey `is set. `validateSignatureValue `will also work with only the signingKey (think about symmetric algorithms)

lib/signed-xml.js
```diff
@@ -359,6 +359,57 @@ SignedXml.defaultNsForPrefix = {
...
+  if (publicCert == null) {
+    return null;
+  }
+
+  prefix = prefix ? prefix + ":" : prefix;
+
+  let x509Certs = "";
+  if (Buffer.isBuffer(publicCert)) {
+    publicCert = publicCert.toString("latin1");
+  }
```
shunkica, line 382: This needs to be documented and typed

test/hmac-tests.js
shunkica, line 1: I think this would be a good test to add.
Currently this is the KeyInfo element generated in this test:
`<KeyInfo><nullX509Data/></KeyInfo>`


```
  it("make sure private hmac key is not leaked due to key confusion", function () {
    var xml = "<library>" + "<book>" + "<name>Harry Potter</name>" + "</book>" + "</library>";
    var sig = new crypto.SignedXml();
    sig.signingKey = fs.readFileSync("./test/static/hmac.key");
    sig.signingCert = fs.readFileSync("./test/static/hmac.key");
    sig.signatureAlgorithm = "http://www.w3.org/2000/09/xmldsig#hmac-sha1";
    sig.addReference("//*[local-name(.)='book']");
    sig.computeSignature(xml);

    var doc = new xmldom.DOMParser().parseFromString(sig.getSignedXml());

    var keyInfo = xpath.select(
        "//*[local-name(.)='KeyInfo']",
        doc
    )[0];

    expect(keyInfo).to.be.undefined;
  });
```

lib/signed-xml.js
```diff
@@ -359,35 +359,65 @@ SignedXml.defaultNsForPrefix = {
...
-  this.signingKey = this.keyInfoProvider.getKey(this.keyInfo);
-  if (!this.signingKey) {
-    var err2 = new Error("key info provider could not resolve key info " + this.keyInfo);
-    if (!callback) {
-      throw err2;
-    } else {
-      callback(err2);
-      return;
+  if (typeof publicCert === "string") {
+    publicCert = publicCert.match(utils.MULTI_PEM_SPLIT);
```
shunkica, line 375: We should make sure that private keys are excluded during this check.
This includes tags:

```
-----BEGIN PRIVATE KEY-----
-----BEGIN RSA PRIVATE KEY-----
-----BEGIN ENCRYPTED PRIVATE KEY-----
```

Basically any tags containing the word PRIVATE.

shunkica: Have you considered using the 'crypto' module to parse private and public keys?

test/hmac-tests.js
cjbarth: This test failed as you wrote it here, but I've modified it and included it. Please have a look. Also note that this required that we unset any existing `getKeyInfoContent` function to make it work properly.

cjbarth: The tests are non-hermetic, so with these changes, the tests are failing... I need to figure this out. Any help would be appreciated.

lib/signed-xml.js
cjbarth: I feel like at some point we have to not worry about people doing the wrong thing with such clear labels, but I'll see what I can do. Want to write some tests for this?

test/hmac-tests.js
shunkica: > The tests are non-hermetic, so with these changes, the tests are failing... I need to figure this out. Any help would be appreciated.

Have you tried making SignatureAlgorithms an instance member?
Is there a good reason for any property to be static?

lib/signed-xml.js
shunkica: > I feel like at some point we have to not worry about people doing the wrong thing with such clear labels

At some point yes, but IMO private key confidentiality is not that point.

I will look at it tomorrow if I find the time.

shunkica:
I have created a [pull request ](https://github.com/cjbarth/xml-crypto/pull/1) addressing some bugs and issues that were discussed.

test/hmac-tests.js
bawolff: > I think `SignatureAlgorithms` shouldn't be static if `enableHMAC()` is no longer static.
> 
> We should also consider [this](https://github.com/node-saml/xml-crypto/security/advisories/GHSA-c27r-x354-4m68) when changing anything related to HMAC. ( I do not understand it, as it lacks description on how exactly that might happen. )
> 
> @bawolff

Its been a long time. From what i remember the issue was that the XML document being verified chose the algorithm. If the XML document said it wanted to use HMAC, then the library would use the public key as the HMAC secret, which would be bad since usually public keys are public knowledge and often even included in the XML document being verified. So someone could take a validly signed document, modify it, switch it to use HMAC as the algorithm, swap the sig for when generated using HMAC with the pub key as the secret, and the old version of the library would accept the modified xml document as valid.

Hope that helps.

index.d.ts
```diff
@@ -278,115 +288,79 @@ export class SignedXml {
...
-   * @return an XML string representation of the contents of a KeyInfo element
+   * With few exceptions;
+   *  - 'posteb' MAY have 'eol', but it is not mandatory.
+   *  - 'preeb' and 'posteb' lines are limited to 64 characters, but
+   *     should not cause any issues in context of PKIX, PKCS and CMS.
    */
-  getKeyInfo(key?: string, prefix?: string): string;
+  PEM_FORMAT_REGEX: RegExp;
+  MULTI_PEM_SPLIT: RegExp;
+  BASE64_REGEX: RegExp;
```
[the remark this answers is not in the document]
shunkica: Other than that everything LGTM.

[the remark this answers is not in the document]
cjbarth: I figured I'd export them so that others could use them if they want. There is nothing wrong with them and they are well named. Any objections?

test/hmac-tests.js
cjbarth: > Is there a good reason for any property to be static?

That is what I was thinking. I can't see any reason. If you turn on HMAC, you should have to construct a new instance if you don't want HMAC support any more; you shouldn't be able to swap.

cjbarth:
Thanks to much work by @shunkica , I think we are ready to merge. Are there any final comments @LoneRifle or @djaqua ? I'll leave this open all weekend to give you a chance to look and then we'll merge it late on Monday.

I really appreciate all of your comments. I feel that we have something here that really cleans up the interface a lot, but which doesn't over-engineer this for use-cases that are unidentified.

index.d.ts
shunkica, line 1: Remove static from sig, hash and canon algorithms, as they are instance members now.

LoneRifle (approved):
Let's get this one through for now, we'll figure out further changes along the way

LoneRifle:
@cjbarth - feel free to merge!

PR #315, opened 2023-06-17
Rename `signingCert` -> `publicCert` and `signingKey` -> `privateKey`

cjbarth:


test/validators/XmlCryptoUtilities/XmlCryptoUtilities/program-repro-misc-validation-and-canon.cs
```diff
@@ -194,7 +194,7 @@ public static void SignXmlFile(string FileName, string SignedFileName, RSA Key)
         SignedXml signedXml = new SignedXml(doc);
 
         // Add the key to the SignedXml document. 
-        signedXml.SigningKey = Key;
+        signedXml.privateKey = Key;
```
LoneRifle, line 197: This is actually a dotnet class, and the property is correctly [SigningKey](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.xml.signedxml.signingkey?view=windowsdesktop-7.0#system-security-cryptography-xml-signedxml-signingkey)

test/validators/XmlCryptoUtilities/XmlCryptoUtilities/utilities.cs
```diff
@@ -30,7 +30,7 @@ static void GetSignature()
             var c = new X509Certificate2(
                 File.ReadAllBytes(@"C:\Program Files\Microsoft WSE\v2.0\Samples\Sample Test Certificates\Client Private.pfx"), "wse2qs");
 
-            signedXml.SigningKey = c.PrivateKey;
+            signedXml.privateKey = c.PrivateKey;
```
LoneRifle, line 33: per earlier comment about this being an in-built dotnet class

LoneRifle (changes requested):
lgtm otherwise. Interesting to know that this package may have been modelled after its dotnet equivalent.

test/validators/XmlCryptoUtilities/XmlCryptoUtilities/program-repro-misc-validation-and-canon.cs
cjbarth: I knew that... I just got over-zealous in the refactor. Thanks for catching this... again, the importance of code review!

cjbarth:
It certainly was built on the .Net mode, but, I think we'd all agree, no one praises the .Net model for its simplicity. What is more, it seems like the interface really hides a lot of complexity under the hood; complexity that we just don't have the resources to develop or maintain.

I think I'm happy with the direction we've taken this and I think it gives us room to move forward with the kinds of use-cases that those using this project are encountering.

PR #316, opened 2023-06-17
Prefer ES6 classes to prototypical inheritance

cjbarth:


cjbarth:
It appears that GitHub's diff tool is really bad in this case. If you pull the branch and use another tool it should be much easier to see that all that was done was to convert from prototypical inheritance to ES6 classes. This will greatly aid in adding to and updating `index.d.ts` and `@typedef` JSDoc directives. It also makes clear which parts belong to the class and which don't. I foresee breaking up `signed-xml.js` into multiple files so that we only have one class per file in a future PR.

LoneRifle (approved):
lgtm. for future reference, experiment with unified/split views and hiding whitespace if the PR diff appears mangled, especially if it is known that the changes are mostly mechanical

cjbarth:
@LoneRifle , that is a good tip. Hiding whitespace and turning off annotations shows how simple this PR really is. Thanks!

PR #318, opened 2023-06-21
Refactor classes into their own files

cjbarth:
This moves some code around so that classes, or closely related classes, are in their own files.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -6,6 +6,7 @@
 /// <reference types="node" />
 
 import { SelectedValue } from "xpath";
+import * as crypto from "crypto";
 
 type CanonicalizationAlgorithmType =
   | "http://www.w3.org/TR/2001/REC-xml-c14n-20010315"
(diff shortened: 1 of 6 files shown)


LoneRifle (commented):
lgtm. Some concerns over class naming: rather than strict TitleCase, I would prefer that we respect the casing for acronyms, ie, RSASHA1 instead of RsaSha1. This not only recognises that we are dealing with an acronym here (ie, Secure Hash Algorithm rather than Sha), but also provides an easier upgrade path.

@djaqua may also want to comment with respect to naming conventions

cjbarth:
It makes it harder to read when multiple acronyms are in a row if we maintain all caps. As for migration, they shouldn't be using these classes anyway, they should be using strings like `http://www.w3.org/2001/04/xmldsig-more#rsa-sha256`, which I might also add, are lower-case and delineated (albeit by a dash instead of a capital).

The following guidelines all say to do it this way:

 - https://www.oreilly.com/library/view/java-8-pocket/9781491901083/ch01.html
 - https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions
   - They make an allowance for two-letter acronyms, which they apply inconsistently in .Net, and which I disagree with
 - https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined
 - https://en.wikipedia.org/wiki/Camel_case#Programming_and_coding

This is a contrary example, but no one likes it (see https://stackoverflow.com/a/45860122/271351)
 - https://github.com/airbnb/javascript#naming--Acronyms-and-Initialisms

Which is easier to read `XmlHttpRequest` or `XMLHTTPRequest`? Since RADAR is also an acronym, should we name a property like `UFORADARID` or `UfoRadarId`?

LoneRifle (approved):
Point taken, no further objections!

PR #319, opened 2023-06-22
Update types

cjbarth:
This PR will update some TypeScript types.


cjbarth:
I would like to get your feedback about just converting this project to TypeScript. We can do it progressively like we did with `node-saml`. Trying to figure out the types and keep them separate from the JavaScript is quite a mental load, and I'm not sure of the benefit of doing so.

LoneRifle (approved):

LoneRifle:
No issues with that, I actually prefer TypeScript. If it helps you be more productive on this repo then go for it!

PR #334, opened 2023-07-07
Ignore unnecessary files in the release

### LINKED ISSUE #332: [ENHANCEMENT]: Remove files, folders not needed on the release
### Is your feature request related to a problem? Please describe...

Why are those folders and files in node_modules?
image

### Describe teh solution you'd like...

remove them

### Describe the alternatives you've considered...

I have not considered any other alternative

angeljqv:
Closes #332

> If you are in a position to help get that fixed for the 3.x branch, a PR is welcome as we are soon to release 3.2.

@cjbarth is it alright???

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,11 +23,8 @@
   "types": "./index.d.ts",
   "files": [
     "CHANGELOG.md",
-    "LICENSE",
-    "README.md",
     "index.d.ts",
     "index.js",
-    "package.json",
     "lib"
   ],
   "directories": {
--- a/package.json
+++ b/package.json
@@ -21,6 +21,15 @@
   ],
   "main": "./index.js",
   "types": "./index.d.ts",
+  "files": [
+    "CHANGELOG.md",
+    "LICENSE",
+    "README.md",
+    "index.d.ts",
+    "index.js",
+    "package.json",
+    "lib"
+  ],
   "directories": {
     "lib": "./lib"
   },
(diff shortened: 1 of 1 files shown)


cjbarth:
That's pretty good, but you have too much there. Please see https://docs.npmjs.com/cli/v9/configuring-npm/package-json#files

parallels999:
Great 👍

package.json
```diff
@@ -24,7 +24,6 @@
   "files": [
     "CHANGELOG.md",
     "index.d.ts",
-    "index.js",
```
parallels999: @cjbarth hi, Is this change correct?

I already read this: [configuring-npm/package-json#files](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#files)
>Certain files are always included, regardless of settings: The file in the "main" field

But with `npm pack` i get this:

image

`index.js` is missing, i'm doing something wrong?? or is `main` failing because it is `./index.js`

parallels999: if you remove `./` it works
https://github.com/node-saml/xml-crypto/blob/6bcbaa6f602aef67f6f5a291b3760ce00b67d997/package.json#L22-L23
Maybe here too
https://github.com/node-saml/xml-crypto/blob/6bcbaa6f602aef67f6f5a291b3760ce00b67d997/package.json#L30

cjbarth: Good find. I was just about to do a release. Please make a PR for this and we'll get it in :) I've been jumping around too much and missed this. Thanks for keeping an eye on this!

PR #361, opened 2023-07-21
Fix validationErrors type

BC-M:
In the examples it is shown that the value can be read. Therefore, the value should not be marked as private.


cjbarth:
Thank you for bringing this to our attention. There are actually a few typing errors like this. I'll review this and see if this fix by itself is sufficient.

cjbarth (approved):

PR #364, opened 2023-07-27
Fix `pemToDer()` return type

cjbarth:
A DER-encoded certificate should be in binary format. This fix is technically a breaking change, but it is also a bug fix. It is unlikely that many are using `pemToDer()` on their own, so the risk of automatic breaking is low with a semver-minor release.


LoneRifle (approved):
Lgtm.

PR #373, opened 2023-08-03
Improve and simplify validation logic

cjbarth:
The validation logic was a little convoluted. This PR aims to clean up that logic a little to make it easier to reason on. The deprecated functions can be removed in the next major release.


LoneRifle (approved):
Lgtm. Nice work untangling the flow between `checkSignature` and `validateSignatureValue`, so that we can remove the latter!

Issue #376, opened 2023-08-13
invalid signature: for uri  calculated digest is '*' but the xml to validate supplies digest '*'


CarlaMck77:
Good Day,

I have seen this reported a few different time with no resolution. I am hoping I would get some assistance in resolving this.
I am new to this type of feature, so any guidance or assistance is more than welcome. I have battling this issue for a few days now and I have no idea what to do next.

I am using version **"xml-crypto": "^4.0.1"**

node version **v16.14.0**

Error displayed as 

**invalid signature: for uri  calculated digest is 2U1suBt1sOA2olbnbMK1gC/3FHk= but the xml to validate supplies digest OGXcEIgUP1W+Hv9ghexl8gdMtrI=**

Generated XML

```
<samlp:Response xmlns:samlp="urn:oasis:names:tc:SAML:1.0:protocol">
    <saml:Assertion MajorVersion="1" MinorVersion="1"
        AssertionID="_1" Issuer="mydomain.com"
        IssueInstant="2023-08-13T03:01:27.265Z" xmlns:saml="urn:oasis:names:tc:SAML:1.0:assertion"
        Id="_0">
        <saml:Conditions NotBefore="2023-08-13T03:01:27.266Z"
            NotOnOrAfter="2023-08-14T03:01:27.263Z" />
        <saml:AuthenticationStatement AuthenticationMethod="urn:oasis:names:tc:SAML:1.0:am:password"
            AuthenticationInstant="2023-08-13T03:01:27.268Z">
            <saml:Subject>
                <saml:NameIdentifier Format="urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified"
                    NameQualifier="urn:mydomain.com">123356</saml:NameIdentifier>
                <saml:SubjectConfirmation>
                    <saml:ConfirmationMethod>urn:oasis:names:tc:SAML:1.0:cm:bearer</saml:ConfirmationMethod>
                </saml:SubjectConfirmation>
            </saml:Subject>
        </saml:AuthenticationStatement>
        <saml:AttributeStatement>
            <saml:Subject>
                <saml:NameIdentifier Format="urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified"
                    NameQualifier="urn:mydomain.com">123356</saml:NameIdentifier>
                <saml:SubjectConfirmation>
                    <saml:ConfirmationMethod>urn:oasis:names:tc:SAML:1.0:cm:bearer</saml:ConfirmationMethod>
[...]

PR #405, opened 2023-10-12
Remove default canonicalization algorithm

cjbarth:
This removes the default canonicalization algorithm forcing the consumer to choose the one that they want. This is an attempt to address #376.

Side note: We do not support all the "required" algorithms. See: https://www.w3.org/TR/xmldsig-core1/#:~:text=4.1%20DSA%5D-,canonicalization,-Required


LoneRifle (approved):
lgtm

PR #406, opened 2023-10-18
Remove default for digest algorithm

cjbarth:
This removes the default digest algorithm forcing the consumer to choose the one that they want. This is an attempt to address https://github.com/node-saml/xml-crypto/issues/376.


LoneRifle (approved):
lgtm

PR #408, opened 2023-10-19
Remove default for signature algorithm

cjbarth:
This removes the default signature algorithm forcing the consumer to choose the one that they want. This is an attempt to address https://github.com/node-saml/xml-crypto/issues/376.

### DIFF
--- a/src/signed-xml.ts
+++ b/src/signed-xml.ts
@@ -37,7 +37,7 @@ export class SignedXml {
    * One of the supported signature algorithms.
    * @see {@link SignatureAlgorithmType}
    */
-  signatureAlgorithm: SignatureAlgorithmType = "http://www.w3.org/2000/09/xmldsig#rsa-sha1";
+  signatureAlgorithm?: SignatureAlgorithmType = undefined;
   /**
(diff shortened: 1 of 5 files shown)


src/signed-xml.ts
```diff
@@ -347,7 +347,10 @@ export class SignedXml {
     }
   }
 
-  private findSignatureAlgorithm(name: SignatureAlgorithmType) {
+  private findSignatureAlgorithm(name?: SignatureAlgorithmType) {
```
LoneRifle, line 350: Given that we now require a name, you may as well enforce the specification of this parameter via types
```suggestion
  private findSignatureAlgorithm(name: SignatureAlgorithmType) {
```

LoneRifle (approved):
lgtm otherwise

src/signed-xml.ts
cjbarth: It was less code changes to allow this to be optional as an argument and throw than it was to add a check at each call-site. If you're prefer to see a throw at each call-site, I can see that logic (allowing for different error messages, potentially easier debugging), and I'll make such a change at your advisement. LMK.

PR #410, opened 2023-11-11
Remove default for transformation algorithm

cjbarth:
This removes the default transformation algorithm, forcing the consumer to choose the one that they want. This is an attempt to address https://github.com/node-saml/xml-crypto/issues/376.

I've also reviewed the README to update a few things that were missed in previous PR removing defaults.


LoneRifle (approved):
lgtm

PR #412, opened 2023-11-25
Mark `getKeyInfo()` private as it has no public consumers

cjbarth:


LoneRifle (approved):

GauriSpears:
Thanks for your kind attitude)
Indeed, getKeyInfoContent wouldn't help because we need to put Object after KeyInfo, not inside it. Actually to satisfy my usecase we need to:
1). add Object node after KeyInfo
and
2). allow Reference to link nodes inside Signature itself (I need to add reference to Signature/Object/QualifyingProperties/SignedProperties).
You are absolutely right, redefining getKeyInfo was rather a dirty hack than a proper way. I'll prepare a PR.

PR #439, opened 2024-01-08
Enhance derToPem to support XML pretty-print

cjbarth:
When pulling this library to `node-saml`, it was discovered that there was a missing case in `derToPem`. This case is not being handled correctly and a test has been added for it.


cjbarth:
@LoneRifle , this should be a simple change to review. It simply removes spaces from a pretty-printed XML payload so that the cert can be properly normalized.

LoneRifle (approved):
lgtm, good catch!

PR #445, opened 2024-01-24
Set `getCertFromKeyInfo` to `noop`

cjbarth:
`getKeyInfoContent` was incorectly set to `noop` instead of `getCertFromKeyInfo`

### DIFF
--- a/src/signed-xml.ts
+++ b/src/signed-xml.ts
@@ -149,8 +149,8 @@ export class SignedXml {
     }
     this.implicitTransforms = implicitTransforms ?? this.implicitTransforms;
     this.keyInfoAttributes = keyInfoAttributes ?? this.keyInfoAttributes;
-    this.getKeyInfoContent = getKeyInfoContent ?? SignedXml.noop;
-    this.getCertFromKeyInfo = getCertFromKeyInfo ?? this.getCertFromKeyInfo;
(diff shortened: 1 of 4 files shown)


cjbarth:
This PR addresses the concerns correctly pointed out by @srd90 [here](https://github.com/node-saml/node-saml/pull/341#issuecomment-1904714894).

Since it is fixing a mistake, and it seems few, if any are using it, should this be a semver-minor change or a semver-major change @LoneRifle ?

LoneRifle (dismissed):
lgtm, thanks for the catch @srd90 !

cjbarth:
@LoneRifle , do you think I should release this as a semver-minor or semver-major release?

LoneRifle:
Inclination is towards a major release, given that it's likely a breaking change. At the same time, you could consider deprecating v5 so that users are encouraged to skip it

cjbarth:
Honestly, I think we de-facto deprecate anything other than the current semver-major release. I suppose I can actively deprecate these on NPM though. I have no interest in trying to support anything other the HEAD or the latest release. If someone wants something on an older version, they can make the correct PR, but that is on them.

@LoneRifle what has been your tradition here?

LoneRifle:
@cjbarth - I've usually just left versions alone,  but only because I previously saw no significant security issues. I'm happy to defer to you to set the norm going forward

Issue #463, opened 2024-04-10
[ENHANCEMENT]: Improve experience of adding a `Reference` to the `Signature`.


cjbarth:
### Is your feature request related to a problem? Please describe...

By default, a `Reference` isn't added for all the first-level nodes in an XML document in the `Signature` block. This means that the `<KeyInfo />` `Reference` isn't included in the `Signature` block.

### Describe teh solution you'd like...

The default, or a simple function call, should be present to allow adding `Reference` elements for all parts of the document being signed.

You can see more about someone trying to do this with C# [here](https://stackoverflow.com/questions/72179983/create-digital-signature-for-xml-with-reference-to-keyinfo).

### Describe the alternatives you've considered...

Adding a `Reference` to each part of the document to be signed is cumbersome and still leaves `<KeyInfo />` without a `Reference`.

PR #464, opened 2024-04-11
Initial code to sign keyinfo

cjbarth:
This is to provide a helper function to sign all children of the root node and a way to include keyinfo in the signature.

Closes: #463

### DIFF
--- a/src/signed-xml.ts
+++ b/src/signed-xml.ts
@@ -654,10 +654,6 @@ export class SignedXml {
           xpath: `./*/*[local-name(.)='${child.localName}']`,
           transforms,
           digestAlgorithm: "http://www.w3.org/2000/09/xmldsig#sha1",
-          // uri: "",
-          // digestValue: "",
-          // inclusiveNamespacesPrefixList: [],
-          // isEmptyUri: true,
         });
       }
     }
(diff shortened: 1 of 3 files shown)


sibelius:
any plans to merge this?

src/signed-xml.ts
```diff
@@ -901,7 +937,14 @@ export class SignedXml {
 
     const keyInfoContent = this.getKeyInfoContent({ publicCert: this.publicCert, prefix });
     if (keyInfoAttrs || keyInfoContent) {
-      return `<${currentPrefix}KeyInfo${keyInfoAttrs}>${keyInfoContent}</${currentPrefix}KeyInfo>`;
+      const keyInfoXml = `<${currentPrefix}KeyInfo${keyInfoAttrs}>${keyInfoContent}<fake>2</fake></${currentPrefix}KeyInfo>`;
```
sibelius, line 940: fake is correct ? 

or just testing ?

sibelius:
should we remove whitespace before signing ?

sibelius:
I'm checking the output from this reference implementation that works https://github.com/bacen/pix-dict-quickstart/blob/master/src/main/java/br/gov/bcb/pi/dict/xml/XMLSigner.java#L63

and the output of this package

xml-crypto does not keep the \n inside <ds:SignatureValue>

here is a print with the signature with \n
image


should I implement or active this somehow ?

Issue #45, opened 2015-03-29
allow to use pem with passphrase


yaronn:
currently if a pem requires a passphrase than the user will be prompt to enter it. effectively this means a passphrase is not supported.

solution is to change signer.sign (also verify?) to accept an object as first parameter: https://nodejs.org/api/crypto.html#crypto_sign_sign_private_key_output_format

this is not available yet in node 0.10.33 (not sure starting which version it is). when used on non supported node version the object parameter to sign will throw:

```
crypto.js:429
  var ret = this._binding.sign(toBuf(key));
                      ^
TypeError: Not a buffer
    at Sign.sign (crypto.js:429:27)
```

tngan:
Just for sharing, now I use this one [digitalbazaar/forge](https://github.com/digitalbazaar/forge) to get the pem protected by a passphrase. They have an API `pki.decryptRsaPrivateKey( file, passphrase )` to decrypt the pem.

bjrmatos:
@tngan cool! i use forge in some of my apps to parse certificates.. we'll have forge in mind when find a solution, thnks

marklagendijk:
I stumbled on the same problem. Full example for the solution using forge:
``` javascript
var signedXml = new SignedXml();
signedXml.signingKey = readPrivateKeyFromProtectedPem('myPrivateKey.pem', 'myPassword');

function readPrivateKeyFromProtectedPem(path, passphrase){
    var pem = fs.readFileSync(path).toString();
    var privateKey = pki.decryptRsaPrivateKey(pem, passphrase);
    return pki.privateKeyToPem(privateKey);
}
```

chrisgh1:
has this changed in recent node versions?  I'm getting the following error which looks like passphrase could be passed in?  But I'm not familiar with how this module works.  

crypto.js:283
  var ret = this._handle.sign(toBuf(key), null, passphrase);

I'm on Node version 6.9.1

cjbarth:
@yaronn , @tngan , would you be interested in creating a PR with a test suite to add that functionality to `xml-crypto`?

cjbarth:
If you'd like to pursue this again, please reopen this.

PR #465, opened 2024-04-16
Add privateKeyPassword option for private key decryption

shunkica:
Introduces the `privateKeyPassword` option, enabling users to define the password for decrypting encrypted private keys.
The option is ignored if `privateKey` is not set, or if it is an instance of `crypto.KeyObject`

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ signature algorithms enabled at same time.
 When signing a xml document you can pass the following options to the `SignedXml` constructor to customize the signature process:
 
 - `privateKey` - **[required]** a `Buffer` or pem encoded `String` containing your private key
-- `privateKeyPassword` - **[optional]** the password to decrypt the private key
+- `privateKeyPassphrase` - **[optional]** the passphrase to decrypt the private key
 - `publicCert` - **[optional]** a `Buffer` or pem encoded `String` containing your public key
 - `signatureAlgorithm` - **[required]** one of the supported [signature algorithms](#signature-algorithms). Ex: `sign.signatureAlgorithm = "http://www.w3.org/2001/04/xmldsig-more#rsa-sha256"`
 - `canonicalizationAlgorithm` - **[required]** one of the supported [canonicalization algorithms](#canonicalization-and-transformation-algorithms). Ex: `sign.canonicalizationAlgorithm = "http://www.w3.org/2001/10/xml-exc-c14n#WithComments"`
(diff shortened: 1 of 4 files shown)


shunkica:
closes #45

README.md
```diff
@@ -56,6 +56,7 @@ signature algorithms enabled at same time.
 When signing a xml document you can pass the following options to the `SignedXml` constructor to customize the signature process:
 
 - `privateKey` - **[required]** a `Buffer` or pem encoded `String` containing your private key
+- `privateKeyPass` - **[optional]** the password to decrypt the private key
```
cjbarth, line 59: Because of a potential confusion between "pass" as in to "pass over" and "pass" as short for "password", please use the longer variable name: `privateKeyPassword`.

src/signed-xml.ts
```diff
@@ -339,13 +340,14 @@ export class SignedXml {
   private calculateSignatureValue(doc: Document, callback?: ErrorFirstCallback<string>) {
     const signedInfoCanon = this.getCanonSignedInfoXml(doc);
     const signer = this.findSignatureAlgorithm(this.signatureAlgorithm);
-    if (this.privateKey == null) {
+    const privateKey = this.getPrivateKey();
+    if (privateKey === undefined) {
```
cjbarth, line 344: Why is `null` now permissible?

test/signature-unit-tests.spec.ts
```diff
@@ -1265,4 +1265,14 @@ describe("Signature unit tests", function () {
       "MIIDZ",
     );
   });
+
+  it("can decrypt the private key when privateKeyPass is set", function () {
```
cjbarth, line 1269: I would be more thorough to also have a test that does throw when trying to use this key when the passphrase isn't provided; please add that.

cjbarth (changes requested):
There are several little things that would up the quality of this PR in measurable ways. Please see the comments.

src/signed-xml.ts
shunkica: `getPrivateKey()` will never return null.
It could never have been null to begin with. It could only be undefined.

cjbarth: That doesn't appear to be true. While the typing system would prevent anything other than those values, not everyone uses Typescript. I see no code that would prevent any value from being used for `privateKey`.

README.md
cjbarth: I've renamed this to `privateKeyPassphrase` for you.

cjbarth:
One more thought for you: do you think it would make sense to allow a `string` or a function that returns `string` for `privateKeyPassphrase`? That would allow a function to retrieve that data from a secure vault or environment variable of some sort.

Thank you for the more complete tests.

src/signed-xml.ts
shunkica: I dont think thats a rabbit hole worth going down.
By that logic `this.privateKey == null` also does nothing, because i could set privateKey=1, privateKey='' etc.., and that check would not catch it.
You cant reasonably expect the library to work if you set privateKey to null.
In any case an error will be thrown by the crypto library.

cjbarth:
@shunkica , I was just looking back at some PRs to see if we could get them landed. Let me know what you think about my last comment.

shunkica:
> @shunkica , I was just looking back at some PRs to see if we could get them landed. Let me know what you think about my last comment.

Sorry this is not something I have bandwidth for right now. If you have time to implement it go ahead.

Issue #328, opened 2023-07-02
[ENHANCEMENT]: Signature compliant to http://www.w3.org/2007/05/xmldsig-more#sha256-rsa-MGF1


rahxam:
### Is your feature request related to a problem? Please describe...

I am trying to connect to a SAML IDP which expects a http://www.w3.org/2007/05/xmldsig-more#sha256-rsa-MGF1 Signature as described in rfc6931 [RSASSA-PSS without Parameters](https://www.rfc-editor.org/rfc/rfc6931.html#section-2.3.10).

Unfortunately, I am a bit stuck on how to implement it.

### Describe teh solution you'd like...

I would like to have a new option to use http://www.w3.org/2007/05/xmldsig-more#sha256-rsa-MGF1 signatures.

### Describe the alternatives you've considered...

I tried to use https://github.com/digitalbazaar/forge to calculate the signature, but did not have any success.

cjbarth:
The first step would be to see if NodeJS supports this. If so, then you can easily put up a PR that mimics the existing methods for doing this. If not, you'll have to figure that out yourself and create a function to do it by hand.

rahxam:
I was able to implement it with https://www.npmjs.com/package/node-forge#rsa, but I guess the dependency is a bit overkill for xml-crypto.

rahxam:
Hey,
Yes, I actually did, but I am using passport-saml and I needed to fork node-saml and xml-encryption as well to use the algorithm in xml-crypto and get everything running, which is lot's of forks for 10 lines of code.

cjbarth:
If you'd like to add support for custom signing methods to `node-saml`, I'd be happy to look at that. This way you could just pass your function through `node-saml` to `xml-crypto`.

cjbarth:
@cornzz , I've been a little busy lately, but I always try to do a big release of all the updates and new features to coincide with the NodeJS release schedule. So, in the coming months we'll be doing that. I'm not writing much code for this anymore. The only bit of code that I'm thinking I might get to is https://github.com/node-saml/xml-crypto/pull/464. Other than that, I'll review code and consult on development, but I depend on others to write it. If you and @rahxam  would like to put up a PR, I'll gladly look at it and get your changes merged in so that you don't have to use a fork.

cornzz:
[...]

PR #488, opened 2025-02-13
Add support for sha256-rsa-MGF1 signing algorithm (#328)

cornzz:
Solves #328

### DIFF
--- a/src/signature-algorithms.ts
+++ b/src/signature-algorithms.ts
@@ -53,6 +53,53 @@ export class RsaSha256 implements SignatureAlgorithm {
   };
 }
 
+export class RsaSha256Mgf1 implements SignatureAlgorithm {
+  getSignature = createOptionalCallbackFunction(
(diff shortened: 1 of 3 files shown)


cjbarth:
Thank you very much for your contribution. Can you please include tests?

cornzz:
Could you give me a hint where to start? Im a bit lost, there are no tests testing each of the signing methods, just for sha1. Also, the result of sha256 with mgf1 signing is not deterministic, so i cant just compare the signature to a hardcoded value as is being done in all other tests. Likewise in `node-saml`. I tried writing a test that uses `validateSignature` to validate the generated signature but its not working...

cjbarth (approved):

cornzz:
@kaibernhard @cjbarth Sorry for not reacting, I completely forgot about this issue and had notifications turned off. Thanks a lot for taking care of the tests! If there is anything else I can do let me know.

PR #489, opened 2025-03-14
Introduce new .getSignedReferences() function of signature to help prevent signature wrapping attacks

ahacker1-securesaml:


CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+## 6.1.0 (...)
+
+# Changelog
+
+- Introduced new .signedReferences property of signature to help prevent signature wrapping attacks.
```
cjbarth, line 7: The changelog is automatically generated from the title of the PR. Any additional notes should go in the README.

src/signed-xml.ts
```diff
@@ -87,6 +88,8 @@ export class SignedXml {
     "http://www.w3.org/2000/09/xmldsig#enveloped-signature": envelopedSignatures.EnvelopedSignature,
   };
 
+  // TODO: In V7.x we may consider deprecating sha1
```
cjbarth, line 91: I'm all for putting in a deprecation warning now and removing it later on. Either way.

src/signed-xml.ts
```diff
@@ -154,6 +160,9 @@ export class SignedXml {
     this.CanonicalizationAlgorithms;
     this.HashAlgorithms;
     this.SignatureAlgorithms;
+    // this populates only after verifying the signature
+    // array of bytes that are cryptographically authenticated
+    this.signedReferences = []; // TODO: should we rename this to something better.
```
cjbarth, line 165: That seems like a descriptive name to me. What other thoughts do you have?

src/signed-xml.ts
```diff
@@ -307,25 +317,45 @@ export class SignedXml {
...
+      if (callback) {
+        callback(null, true);
+      } else {
+        return true;
+      }
     } else {
-      const verified = signer.verifySignature(unverifiedSignedInfoCanon, key, this.signatureValue);
+      // false case
+      // reset the signedReferences back to empty array
+      // I would have preferred to start by verifying the signedInfoCanon first, if that's OK
```
cjbarth, line 344: I feel like our tests would catch if this causes breaking changes. Did one of the existing tests fail?

cjbarth (changes requested):
This looks good, but might need a lint run over it. I can do that just before I land it.

I'm marking this PR as "Request Changes" so you can see the comments on the PR and at least acknowledge them (make resolved if you want). They aren't major.

src/signed-xml.ts
cjbarth: Sure. I was thinking there is no reason not to throw a deprecation notice now if someone uses sha1, no need to wait for 7.x to throw the notice.

src/signed-xml.ts
cjbarth: Chatty suggested options like `authenticatedData` and `signatureVerifiedBytes`. I have no problems with log descriptive names. I think, at a minimum, we should make sure the type is correct; right now it is untyped.

src/signed-xml.ts
cjbarth: If all tests past when verifying `signedInfoCanon` first, I'm ok if you wanted to make that change now. If you still think it is too risky, I understand, though I'd be interested in what test might give us reassurance.

cjbarth:
@ahacker1-securesaml , if the changes I've just pushed look good, I'll land this and release.

cjbarth:
I've tried to put up some edits to the README, but it appears that you've restricted my ability to push to this branch. Was that intentional @ahacker1-securesaml ?

cjbarth:
@ahacker1-securesaml , I've pushed up some changes. I'd like to get at least one test for this. I've been trying to inspect the other tests and I haven't actually found a case where `.signedReferences` is populated. Do you happen to have a test? That would speed this up. Even just the rough structure or sample code you used for local testing. I can make it fit with this project.

src/signed-xml.ts
```diff
@@ -299,6 +308,7 @@ export class SignedXml {
     }
 
     if (!this.getReferences().every((ref) => this.validateReference(ref, doc))) {
+      this.signedReferences = []; // reset again
```
cjbarth, line 311: Why do we have to reset this here? Shouldn't the `.signedReferences` array contain the validated references instead of no references?

If we are going to reset this array, then v7.x should probably just leave `getReferences()` alone and just throw if the references don't validate. In that case, we don't even need a 6.1 release because it would be a no-code-change-just-behavior-change to throw here and leave everything else alone.

I have a branch where I've updated tests to account for `signedReferences` being a different length than `getReferences()`. I want to discuss this before I push that up over what you have.

src/signed-xml.ts
```diff
@@ -307,25 +317,40 @@ export class SignedXml {
...
+    const sigRes = signer.verifySignature(unverifiedSignedInfoCanon, key, this.signatureValue);
+    if (sigRes === true) {
+      if (callback) {
+        callback(null, true);
+      } else {
+        return true;
+      }
     } else {
-      const verified = signer.verifySignature(unverifiedSignedInfoCanon, key, this.signatureValue);
+      // Reset the signedReferences back to empty array.
```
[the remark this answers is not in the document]
cjbarth: Note made. I've updated a comment in my working branch to reflect this.

src/signed-xml.ts
```diff
@@ -307,25 +316,40 @@ export class SignedXml {
...
+      // Ideally we would start by verifying the signedInfoCanon first,
+      // but that may cause some breaking changes, so we'll handle that in v7.x
+      this.signedReferences = [];
```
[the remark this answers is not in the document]
cjbarth: @srd90 , I noted this too when I was working on some tests, but I commented before I saw this here: https://github.com/node-saml/xml-crypto/pull/489#discussion_r2032014902

Thoughts?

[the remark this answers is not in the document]
cjbarth: @ahacker1-securesaml , if the goal is to have `.signedReferences` eventually contain the exact bytes there were validated, then why are we putting in the XML that we validated in this version? Wouldn't that cause even more code changes down the line?

markstos:
> However, I have deadlines to meet. The code currently just works with other SAML libraries. We want to ship it ASAP.

Appreciate the collaboration here, but the bottom line priority here is security of node-saml, unconstrained by some organization's internal deadlines.

src/signed-xml.ts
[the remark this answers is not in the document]
cjbarth: @ahacker1-securesaml , I am trying to get this resolve quickly. As I said, I have a whole branch with a bunch of tests on it. But, if we say the code contains the exact bytes, then we should probably do that. Please note what `.signedReferences` actually contains upon debug inspection.

image

src/signed-xml.ts
cjbarth: Your point about accessing the signed, authenticated, bytes is valid. Of course, per my previous comment, that isn't what `.signedReferences` contains, but that is another matter.

I want to reemphasize, I'm not against releasing 6.1. I also very much appreciate your work and collaboration. Please keep in mind that this isn't my day job, so I'm doing the best I can to keep code quality high and to help the community here.

My question wasn't about pushing off this change or changing the API, but rather, "why are we resetting this `.signedReferences` here in the case that a single reference fails to validate?" Is there value in leaving `.signedReferences` populated with the references that did validate?

src/signed-xml.ts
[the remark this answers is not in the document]
cjbarth: Thank you for that, but it doesn't track because bytes are _never_ encoded as strings (rendered, yes, but not encoded), strings are encoded as bytes, thus my confusion. I refreshed my memory with some research and found that the canonicalization process always results in utf-8 encoded XML strings, so the signature is always calculated against an XML node encoded to bytes as utf-8.

So, while you keep saying bytes, what you're actually referring to is the canonical XML node that was validly signed, which, incidentally, happens to be stored in `.signedReferences` as a utf-8 encoded string (thus the exact same bytes) because that is how Node natively deals with strings, but is ultimately language dependent. You can't just send that string to another system and guarantee that the bytes didn't get changed to match the receiving systems encoding because you are dealing with strings and system will mutate strings for compatibility.

So, I understand now what you are saying, but, depending on your frame of reference, it can be very misleading; bytes are always bytes; strings are a human interpretation of the data stored in those bytes. See [Mojibake](https://en.wikipedia.org/wiki/Mojibake).

src/signed-xml.ts
cjbarth: Thank you @ahacker1-securesaml . We've always asked for specifications to make sure that we're getting this right and preserving it for the future. In this day and age of people poisoning libraries it is wise for none of us to blindly listen to someone say "I'm awesome, trust me". [Trust, but verify](https://en.wikipedia.org/wiki/Trust,_but_verify).

Thus, in that spirit, I did the research myself and found the specification you must be referring to whereby we've been validating references wrong: https://www.w3.org/TR/xmldsig-core/#sec-CoreValidation

I will update the comments in the code to show this so that we make sure to fix it (breaking change) in v7.x.

cjbarth:
@ahacker1-securesaml  and @srd90 , have a look at the latest code I pushed. It is linted and tested and has updated comments to reflect our discussion here. I'm content to land this and release it as 6.1 if you both agree.

test/document-tests.spec.ts
```diff
@@ -107,11 +116,14 @@ describe("Validated node references tests", function () {
...
     expect(validSignature).to.be.false;
+    expect(sig.signedReferences.length).to.equal(0);
 
     const ref = sig.getReferences()[0];
     const result = ref.getValidatedNode(
       "//*[local-name()='Attribute' and @Name='mail']/*[local-name()='AttributeValue']/text()",
     );
     expect(result).to.be.null;
+    // need all references to verify. Reset back to zero now
+    expect(sig.signedReferences.length).to.equal(0);
```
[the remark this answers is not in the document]
cjbarth: `result` is taken from only the first reference returned by `getReferences()`. The check here will be to make sure that there are no `signedReferences` at all.

cjbarth:
@ahacker1-securesaml , just one more thought. We have `getReferences()` as a function that simply returns an array to make sure that no consumer of the library can write to that array. Would a similar pattern be appropriate here to make sure that a consumer can't accidentally or intentionally modify `.signedReferences` themselves? We could make it an internal array and then expose `getSignedReferences()`. Such a refactor would only take a moment on my end. To make doubly sure, I'd probably return `[...this.signedReferences]` to get a copy, which is even better than what we do now.

src/signed-xml.ts
cjbarth: Thank you for that. In the case that we're dealing with, the data object is a canonicalized XML element, which, due to the rules of canonicalization, is the XML element as a string, encoded to bytes as utf-8. So, it is just super convenient that Node also treats strings as utf-8, so we don't have to work in bytes, we can work in strings, and it all works out. Python 2 would be an entirely different story.

src/signed-xml.ts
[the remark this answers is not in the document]
cjbarth: I'm fine with this. I've updated the comments to reflect what the user will see.

cjbarth:
@ahacker1-securesaml  and @srd90 , are we all good now?

cjbarth:
@ahacker1-securesaml or @srd90 , just so I make sure I'm not missing anything, can you comment on the subtle differences between the new `.getSignedReferences()` API and the previously recommended technique of using `.validateElementAgainstReferences()` API. It seems the former is a list of the validated references and the latter is a function to take any xpath, look it up in the document, and see if that specific element passes validation.

I'm not saying that we shouldn't have the `.getSignedReferences()` API, but I'm wondering if we should also deprecate `.validateElementAgainstReference()` API. I'm sure that API was written specifically to help people validate references, though, admittedly, in a not very elegant way.

I'm trying to clean up all the internal uses of `getReferences()` so that we don't have any internal deprecation warnings, and I want to make sure I get this 100% correct.

cjbarth:
I'll turn my attention back to `node-saml` then.

So then, `getReferences()` shouldn't have been deprecated?

cjbarth:
I'm not sure I fully understand. If the `.reference` contains non-canonical data that is potentially dangerous, why would we say they can use it at all? Are you referring to the other properties that are returned? If so, we can add a test to make sure there is a 1-1 correspondence between `getReferences()` and `getSignedReferences()` and then deprecate only the property of `getReferences()` that returns the non-canonical reference. In fact, we can actually just make the v7 simply change the value of the property to be the same as is in `getSignedReferences()`.

Or, I could see a very strong argument for doing that in a minor release because what we're returning now (the non-canonical data) is dangerous and wrong, but the canonical data would be a minor, and typically non-consequential change for most users.

Even better would be to deprecate the `references[].getValidatedNode()` and add a `references[].getSignedReference()` property that would just return the canonicalized data, the same data that would be found in `getSignedReferences()`. That would be a very simple change that wouldn't break anyone and would reduce the instances of dealing with deprecations that some users are facing.

What am I missing here?

cjbarth:
Ok, and the _only_ reason I care about this library is for SAML applications, like `node-saml`. I don't want to ship `node-saml` where during its normal, expected, runtime operations, it throws deprecation warnings. That is the standard that should be expected. It can't be deprecated if there is no alternative.

So, for the sake of anything related to this discussion, we can put a note at the top of the README directing people to use libsodium, but we have to figure out a way to not have deprecation warnings with normally running secure-as-we-can-make-it-right-now code.

PR #491, opened 2025-03-18
Add .signedReferences API

ahacker1-securesaml:
cc @cjbarth 

Unfortunately, I've decided that the best way is to backport the API. 
Because we still want other libraries to use the API even if they are still on v2.x/v3.x

It is actually a much more quicker change than the libraries upgrading to say v6.x or v7.x


cjbarth:
@ahacker1-securesaml , it may be quicker for you, but it is extra maintenance burden for me and sends the message that we still support those older versions. I'm not going to not do merges and releases, but I'm not in favor of it.

What happened to the idea of breaking changes in a 7.x release that would remove the foot guns and generally clean up the API as you previously proposed?

ahacker1-securesaml:
@coderabbitai hi

Issue #108, opened 2016-07-05
Break the signature validation (>0.8.2)


tngan:
It works in release 0.8.2 but not 0.8.3. Follow the steps to replicate this issue.

**This is the test script, test.js**

``` javascript
var SignedXml = require('xml-crypto').SignedXml;
var dom = require('xmldom').DOMParser;
var rfc1951 = require('deflate-js');
var forge = require('node-forge');
var pki = forge.pki;
var fs = require('fs');
var select = require('xml-crypto').xpath
var dom = require('xmldom').DOMParser
var FileKeyInfo = require('xml-crypto').FileKeyInfo  

function  getKeyInfo(x509Certificate) {
  this.getKeyInfo = function(key) {
    return '<X509Data><X509Certificate>' + x509Certificate + '</X509Certificate></X509Data>';
  };
  this.getKey = function(keyInfo) {
      return getPublicKeyPemFromCertificate(x509Certificate).toString();
  };
}

function getPublicKeyPemFromCertificate(x509Certificate){
    var certDerBytes = forge.util.decode64(x509Certificate);
    var obj = forge.asn1.fromDer(certDerBytes);
    var cert = forge.pki.certificateFromAsn1(obj);
    return pki.publicKeyToPem(cert.publicKey);
}

var xml = fs.readFileSync("signed.xml").toString()
var doc = new dom().parseFromString(xml)    

var signature = select(doc, "/*/*[local-name(.)='Signature' and namespace-uri(.)='http://www.w3.org/2000/09/xmldsig#']")[0]
var sig = new SignedXml()
 //sig.keyInfoProvider = new FileKeyInfo("client_public.pem")
[...]

PR #492, opened 2025-04-07
Accept DOM node when computing/verifying signatures

elbuo8:
In order to handle different versions `@xmldom/xmldom` from the upstream vs `xml-crypto` it would be ideal to support injecting the DOM similarly to how other methods in the library already allow it.

This would allow a faster fixing cycle when cases as these happen: https://github.com/node-saml/xml-crypto/issues/108

Happy to do any additional modifications to improve this so it can be landed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,7 +287,8 @@ To verify xml documents:
 
 - `loadSignature(signatureXml)` - loads the signature where:
   - `signatureXml` - a string or node object (like an [xmldom](https://github.com/xmldom/xmldom) node) containing the xml representation of the signature
-- `checkSignature(xml)` - validates the given xml document and returns `true` if the validation was successful
+- `checkSignature(xml)` - validates the given xml document and returns `true` if the validation was successful where:
+  - `xml` - a string or node object (like an [xmldom](https://github.com/xmldom/xmldom) node) containing the xml representation of the document
 
 ## Customizing Algorithms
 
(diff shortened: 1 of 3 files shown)


cjbarth:
We're in an active update cycle for this library now, so your changes are coming at a good time. Do any of the recent changes affect what you're trying to accomplish?

adamjmcgrath (approved):

elbuo8:
> Do any of the recent changes affect what you're trying to accomplish?

The recent addition/changes of `signedReferences` works fine. We are tracking the V7 Draft but it still not clear how the interface/methods are going to end up. Ideally (and we'd be more than happy to contribute) they would accept DOMs as we are suggesting here so that the versions of `xmldom` are completely de-coupled, if possible.

elbuo8:
While I'd definitely prefer to fix the issue at the downstream dependency, that cycle would take far longer and furthermore, it wouldn't prevent a similar issue in the future given that the upstream can still be using a completely different version of xmldom (which may have other fixes and/or regressions) that do not play well with the version shipped with xml-crypto.

Happy to PR both versions. But I might need some guidance on V7 as I'm seeing `checkSignature` become `_checkSignature` but then not clear which is ultimately the exposed/public API that will be the entrypoint.

elbuo8:
Btw, for the specific bug I'm hunting (and the ticket refers) it's already fixed in xmldom 0.9.

PR #493, opened 2025-04-08
DRAFT: v7.x

ahacker1-securesaml:
This significantly improves the security logic around xml-crypto.

Currently, it is an outline, i.e. the code doesn't actually work

Reviewers:
- Let me know about the security of this, does it make it easier to read, so more secure?
- is there a better way to design this? 
Currently we remove the checkSignature method from the SignedXML class to simplify the logic, and put it in XMLVerifier

Release process debate:
- Do we even want to release 7.x
If feel like the best we can do is to deprecate the library entirely,
because xml signatures as a whole is unsafe, and using it will ensure the other side will be insecure. (since other implementations are insecure)


cjbarth:
Given the number of people using xml-crypto and the lack of an alternative (that I'm aware of). It seems prudent to continue developing this. Otherwise, we should provide a migration guide to some other method of doing what this library does.

PR #500, opened 2025-04-22
Remove all reference XML data if any are corrupted

cjbarth:


markstos (commented):
Makes sense. Given this change, do we still need the TODO comments, or does this change address that goal?

>   // TODO: add this breaking change here later on for even more security: `this.references = [];`

markstos (approved):
Makes sense. In that case, approved as-is.

Issue #486, opened 2025-02-04
Difference from javax.xml.crypto.dsig and xml-crypto signature value


sibelius:
This is the SignedInfo canonilized

```
<ds:SignedInfo xmlns:ds="http://www.w3.org/2000/09/xmldsig#"><ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"></ds:CanonicalizationMethod><ds:SignatureMethod Algorithm="http://www.w3.org/2001/04/xmldsig-more#rsa-sha256"></ds:SignatureMethod><ds:Reference URI="key-info-id"><ds:Transforms><ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"></ds:Transform></ds:Transforms><ds:DigestMethod Algorithm="http://www.w3.org/2001/04/xmlenc#sha256"></ds:DigestMethod><ds:DigestValue>aUK42Eje/aj9EEEZn1wHoSsQMnPW0J8GpDP124YLE1w=</ds:DigestValue></ds:Reference><ds:Reference URI=""><ds:Transforms><ds:Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"></ds:Transform><ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"></ds:Transform></ds:Transforms><ds:DigestMethod Algorithm="http://www.w3.org/2001/04/xmlenc#sha256"></ds:DigestMethod><ds:DigestValue>3lnGNn7gACFoylCvc8RXyyg3etLlVjfuxHYcOZhTL+c=</ds:DigestValue></ds:Reference><ds:Reference><ds:Transforms><ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"></ds:Transform></ds:Transforms><ds:DigestMethod Algorithm="http://www.w3.org/2001/04/xmlenc#sha256"></ds:DigestMethod><ds:DigestValue>aIc0+7BXP+Ee09zng0JE21rkzgVlYaa6qMGHIBQKvTc=</ds:DigestValue></ds:Reference></ds:SignedInfo>
```

it generates this hash in node

```
[...]

Issue #484, opened 2025-01-30
Failed to generate a correct signature


hsan8:
Can someone help me to generate same signature like this in nodejs using xml-crypto:
I have two keys :

- **privatekey.pem**  
- **publickey.pem**

```xml
<Req
    xmlns="http://www.abc.com/abc/schema/">
    <Header requestId="f359718a-b759-4617-aebb-1260d98fef3e" timestamp="2025-01-19 15:26:28.552" ver="1.0"/>
    <Signature
        xmlns="http://www.w3.org/2000/09/xmldsig#">
        <SignedInfo>
            <CanonicalizationMethod Algorithm="http://www.w3.org/TR/2001/REC-xml-c14n-20010315"/>
            <SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/>
            <Reference URI="">
                <Transforms>
                    <Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/>
                </Transforms>
                <DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/>
                <DigestValue>RX4gMUYcjhYtNPANxPJ2aCmxXzc=</DigestValue>
            </Reference>
        </SignedInfo>
        <SignatureValue>fptL80Jl6614.....</SignatureValue>
        <KeyInfo>
            <KeyValue>
                <RSAKeyValue>
                    <Modulus>hPN+z6tI/WSPV1sKDDfw2.....</Modulus>
                    <Exponent>AQAB</Exponent>
                </RSAKeyValue>
            </KeyValue>
        </KeyInfo>
    </Signature>
</Req>
```
This is my try in js

```javascript
const crypto = require('crypto');
const SignedXml = require('xml-crypto').SignedXml;
const fs = require('fs').promises;

class XmlSigner {
  constructor() {
    this.XML_TEMPLATE = `<Req
    xmlns="http://www.abc.com/abc/schema/">
    <Header requestId="f359718a-b759-4617-aebb-1260d98fef3e" timestamp="2025-01-19 15:26:28.552" ver="1.0"/></Req>`;
  }

  async loadPrivateKey(pemContent) {
    return pemContent.replace('-----BEGIN PRIVATE KEY-----', '').replace('-----END PRIVATE KEY-----', '').replace(/\s/g, '');
  }

  async loadPublicKey(pemContent) {
[...]

Issue #212, opened 2020-05-20
Wrong Digest Value generated. Actual XML and Expected XML is not matching. Working from last three nights.


niladrinic:
Dear All, 
   I have used this library to generate a signature which is done at root level. The signed xml is not matching as per the expected xml. The digested value and signature value is not matching with real real value. Can't figure this out working from last three night and almost gone mad. please help me out.

```javascript
var builder = require('xmlbuilder');
var SignedXml = require('xml-crypto').SignedXml;
var fs = require("fs");

module.exports.SXML = function() {
    //Created the request XML as per the required format
    var xml = builder.create('Esign').att('AuthMode', '1').att('aspId', 'TNIC-001').att('kycId', '')
        .att('ekycIdType', 'A').att('responseSigType', "pkcs7")
        .att('responseUrl', 'https://nic-esigngateway.nic.in/eSign21/response?rs=localhost:8080/Esign_client/hello1')
        .att('sc', 'Y').att('ts', '2020-05-19T11:28:01').att('txn', '999-ESIGN-DIVISION-2020-05-19T11:28:01')
        .att('ver', '2.1');
    //xml.ele("Docs").ele("InputHash",{'docInfo':'hello', 'hashAlgorithm':'SHA256', 'id':'1'},'5df2936623312f38371871bf5226e666c66318329d826bf7db544237096a837e');
    xml.ele("Docs").ele("InputHash", {
        'docInfo': 'hello',
        'hashAlgorithm': 'SHA256',
        'id': '1'
    }, '5df2936623312f38371871bf5226e666c66318329d826bf7db544237096a837e');

    // XML doc is ready for signing. Converted to String 
    var xmldoc = xml.toString({
        pretty: true
    });

    var sig = new SignedXml();
    //sig.addReference("/*",['http://www.w3.org/2000/09/xmldsig#enveloped-signature'],'http://www.w3.org/2001/04/xmlenc#sha256','','','',true);
    sig.addReference("/*", ['http://www.w3.org/2000/09/xmldsig#enveloped-signature'], 'http://www.w3.org/2000/09/xmldsig#sha1', '', '', '', true);


    sig.signingKey = fs.readFileSync("nicesign.pem");
    sig.canonicalizationAlgorithm = "http://www.w3.org/TR/2001/REC-xml-c14n-20010315";
    sig.signatureAlgorithm = "http://www.w3.org/2000/09/xmldsig#rsa-sha1";
    sig.computeSignature(xmldoc.toString());

[...]

Issue #210, opened 2020-04-27
enveloped signature transform without c14n-20010315 workaround


Flyingpeanut:
Hi,
First of all I would like to apologize if my issue is redundant, I am not very experienced in xml signatures.
I would like to sign my document with only enveloped-signature as transform
I have based my code on #93 solution
I use this function to sign
```
const forge = require('node-forge'),
    pki = forge.pki,
    select = require('xml-crypto').xpath
  , dom = require('xmldom').DOMParser
  , SignedXml = require('xml-crypto').SignedXml
  , FileKeyInfo = require('xml-crypto').FileKeyInfo  
  , fs = require('fs');

function signXML(){
    let sig = new SignedXml()
  sig.addReference("//*[local-name(.)='elmo']",[
      'http://www.w3.org/2000/09/xmldsig#enveloped-signature',]
      //'http://www.w3.org/TR/2001/REC-xml-c14n-20010315'
     ,'http://www.w3.org/2001/04/xmlenc#sha256',"","","",true)
  sig.signingKey = RSA_key
  sig.keyInfoProvider = new KeyInfoProvider(RSA_cert)
  sig.canonicalizationAlgorithm = 'http://www.w3.org/TR/2001/REC-xml-c14n-20010315';
  sig.signatureAlgorithm = "http://www.w3.org/2000/09/xmldsig#rsa-sha1";
  sig.computeSignature(xml )
  fs.writeFileSync("signed.xml", sig.getSignedXml())
}
// and this to validate the signed xml

function testValidate(){
  var xml = fs.readFileSync("signed.xml").toString()
  var doc = new dom().parseFromString(xml)    
  var signature = select(doc, "//*[local-name(.)='Signature' and namespace-uri(.)='http://www.w3.org/2000/09/xmldsig#']")
  var sig = new SignedXml()
  sig.keyInfoProvider = new KeyInfoProvider(fs.readFileSync("client_public.pem"))
  sig.loadSignature(signature[0].toString())
  var res = sig.checkSignature(xml)
  if (!res) console.log(sig.validationErrors)
    console.log(res);
  return res
}
```
It shouldn't be relevant I am have taken this from another issue #47 to create the X509Data  
```
  function KeyInfoProvider(certificatePEM) {
    if (!this instanceof KeyInfoProvider) {
      return new KeyInfoProvider();
    }
  
    if (Buffer.isBuffer(certificatePEM)) {
      certificatePEM = certificatePEM.toString('ascii');
    }
  
[...]

PR #503, opened 2025-05-02
Add implicit C14N when signing if no transforms specified

roc13x:
When signing, the library currently does not apply any canonicalization to the referenced document being signed, unless explicitly specified in the `transforms` config property.

This behaviour differs from many other XML signature libraries such as javax.xml.crypto.dsig and signxml, who implicitly apply `http://www.w3.org/TR/2001/REC-xml-c14n-20010315` if the user does not specify any canonicalization transforms.

This PR proposes adding the same implicit canonicalization behaviour, to make the output more consistent with other implementations.

Related issues: https://github.com/node-saml/xml-crypto/issues/210 https://github.com/node-saml/xml-crypto/issues/212 https://github.com/node-saml/xml-crypto/issues/484 https://github.com/node-saml/xml-crypto/issues/486

### DIFF
--- a/src/signed-xml.ts
+++ b/src/signed-xml.ts
@@ -423,10 +423,13 @@ export class SignedXml {
   }
 
   private getCanonReferenceXml(doc: Document, ref: Reference, node: Node) {
+    const transforms: string[] = [];
+    
     /**
      * Search for ancestor namespaces before canonicalization.
      */
     if (Array.isArray(ref.transforms)) {
+      transforms.push(...ref.transforms);
       ref.ancestorNamespaces = utils.findAncestorNs(doc, ref.xpath, this.namespaceResolver);
     }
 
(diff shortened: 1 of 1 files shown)


markstos:
It seems the SAML spec does support this as an option:

> http://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf at line 2941:
>
>   Signatures in SAML messages SHOULD NOT contain transforms other than the enveloped 
> signature    transform (with the identifier http://www.w3.org/2000/09/xmldsig#enveloped-signature) > or the exclusive canonicalization transforms (with the identifier http://www.w3.org/2001/10/xml-exc-> c14n# or  http://www.w3.org/2001/10/xml-exc-c14n#WithComments).

My concern is that while this may improve things for some users, it will break them for others, as it changes a default behavior.  There are reports-- I'm not sure if they are current-- that ADFS may have problems and some Java-based servers may have problems as well.

src/signed-xml.ts
```diff
@@ -423,19 +423,29 @@ export class SignedXml {
...
     const c14nOptions = {
       inclusiveNamespacesPrefixList: ref.inclusiveNamespacesPrefixList,
       ancestorNamespaces: ref.ancestorNamespaces,
     };
+    
+    if (
+      transforms.length === 0 ||
+      transforms[transforms.length - 1] === "http://www.w3.org/2000/09/xmldsig#enveloped-signature"
+    ) {
+      transforms.push("http://www.w3.org/TR/2001/REC-xml-c14n-20010315");
```
markstos, line 445: Wouldn't exclusive C14N be more compatible?

```
transforms.push("http://www.w3.org/2001/10/xml-exc-c14n#");
```

cjbarth:
@roc13x, it seems that some of the discussion here indicates that your PR isn't moving in the most preferred direction. We'd like to hear your thoughts if you're still interested in this work being completed. If not, we'll have this PR closed.

cjbarth:
@ahacker1-securesaml , based on your feedback and the comments from @srd90 , we'll close this PR in favor of a different solution in the future.

PR #508, opened 2025-07-12
Respect idAttribute when generating signatures.

### LINKED ISSUE #33: Use this.idAttributes[0] when setting id attribute of element?
I noticed [here](https://github.com/yaronn/xml-crypto/blob/master/lib/signed-xml.js#L483) that the id attribute for elements is set. Can we change the hard-coded `"Id"` to `this.idAttributes[0]`? That way if one passes in `idAttribute` as an option to the SignedXml constructor, it will use that instead.

Basically, I'd like to be able to control whether "Id" or "ID" is outputted when I call SignedXml.computeSignature().

(I'm not familiar with xml-crypto, so my apologies if there's a good reason why it's not this way already.)

tkalmar:
Fixes #33
small fix to respect user provided idAttribute when generating id's

## Summary by CodeRabbit

* **New Features**
  * Support for namespaced ID attributes (e.g., WS‑Security and xml:id) when signing XML.
  * idAttribute option now accepts either a plain name or a namespaced identifier, so you can provide a custom attribute or a namespaced ID.
  * Generated IDs and reference lookups honor the provided attribute form (plain or namespaced) for consistent signing and verification.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,0 +1,18 @@
+import { NamespacedId } from "./types";
+
+/** Well known namespaced ids */
+export const KNOWN_NAMESPACED_IDS: { [key: string]: NamespacedId } = {
+  /** WS-Security */
+  wssecurity: {
+    prefix: "wsu",
+    localName: "Id",
+    nameSpaceURI:
+      "http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-utility-1.0.xsd",
+  },
+  /** Xml */
+  xml: {
+    prefix: "xml",
+    localName: "id",
+    nameSpaceURI: "http://www.w3.org/XML/1998/namespace",
+  },
+};
--- a/src/signed-xml.ts
+++ b/src/signed-xml.ts
@@ -8,6 +8,7 @@ import type {
   GetKeyInfoContentArgs,
   HashAlgorithm,
   HashAlgorithmType,
+  NamespacedId,
   ObjectAttributes,
   Reference,
   SignatureAlgorithm,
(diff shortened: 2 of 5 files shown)


tkalmar:
Not sure if this needs enhancements to provide support for namespaced id's but then the whole idAttribute has to be changed. But in the long ran this can then make the idMode parameter obsolete ...

cjbarth:
So, you don't want a different ID, you want a namespace. Do you have a sample XML that you could share? A proper fix would probably be to read the namespace from the existing XML and use that.

shunkica:
@tkalmar   The XMLDsig spec doesn't prescribe using xml:id specifically—it just requires Id-type attributes. We understand your use case, but I think If there are going to be changes to this it should be to support fully qualified ID attributes.  A simple string-based approach is too narrow and doesn't guarantee proper namespace-aware attribute creation.

  Proposal: Modify this PR to accept idAttribute as either:
  - A string (for backwards compatibility with simple attributes like "Id" or "id")
  - An object: {prefix: string, localName: string, namespaceURI: string}

  For your use case, you'd pass:
```
  {
    prefix: "xml",
    localName: "id",
    namespaceURI: "http://www.w3.org/XML/1998/namespace"
  }
```

  This would properly create a namespaced attribute and accommodate any ID attribute convention—whether it's xml:id,
   wsu:Id, ds:Id, or custom application-specific ones.

tkalmar:
@shunkica A first draft is done.

shunkica:
This does conflict with PR #519 where a functionality like this is already supported.

tkalmar:
@shunkica @cjbarth 
How to proceed from here?
Wait until PR #519 is merged?
Revert to the initial solution?
In my opinion it would be best to revert this to the initial solution.
It fixes exactly one problem.
It is minimal invasive. No breakage no changed params.
It is sufficient for my use case.
WDYT?

cjbarth:
@tkalmar , @shunkica  has been active on #519 and we're about to land it. Would you like to look at it and see if it will suite your needs? Otherwise, please comment on the direction you'd like to see this go since there will be merge conflicts.

Issue #477, opened 2024-09-17
[ENHANCEMENT]: web-crypto support


markusahlstrand:
### Is your feature request related to a problem? Please describe...

I would like to use xml-crypto in WinterCG environments like cloudflare workers.

### Describe teh solution you'd like...

Replace the dependency on node crypto with web-crypto. It should be backwards compatible to node 16.

### Describe the alternatives you've considered...

There seems to be no easy way to shim the node-crypto library. As web-crypto async it also requires the interfaces to update which makes the changes needed fairly large.

ngearhart:
I am working on this

markusahlstrand:
It looks like it could be done without to many changes by using the callback and implementing a new hash-algorithm and signature-algorihm. Maybe this is good grunt work for Claude :)

PR #513, opened 2025-10-06
Added web-crypto support (#1)

markusahlstrand:
# Add WebCrypto API Support for Browser Compatibility

## Overview
This PR adds comprehensive WebCrypto API support to **xml-crypto**, enabling the library to work in browser environments and modern JavaScript runtimes that lack Node.js’s `crypto` module.  
It maintains full backward compatibility with existing Node.js code while introducing a modern, standards-based cryptographic API.

## Motivation
The current implementation relies exclusively on Node.js’s `crypto` module, preventing use in:

- Browser environments  
- Edge computing platforms (Cloudflare Workers, Deno Deploy, etc.)  
- Modern JavaScript runtimes that favor Web standards  

Adding WebCrypto support enables **xml-crypto** to run in all JavaScript environments while maintaining the same API surface.

---

### 1. Dual Cryptography Backend
- Automatic runtime detection between Node.js `crypto` and WebCrypto  
- Seamless fallback mechanism for maximum compatibility  
- Zero breaking changes to existing code  

### 2. New WebCrypto Signature Algorithms (`signature-algorithms-webcrypto.ts`)
- `RsaSha1WebCrypto` – RSA-SHA1 signatures  
- `RsaSha256WebCrypto` – RSA-SHA256 signatures  
- `RsaSha512WebCrypto` – RSA-SHA512 signatures  
- `HmacSha1WebCrypto` – HMAC-SHA1 signatures  
- `HmacSha256WebCrypto` – HMAC-SHA256 signatures  

**All implementations support:**
- Async signing and verification via `signAsync()` and `verifySignatureAsync()`  
- Multiple key formats: `CryptoKey`, `KeyObject`, `Buffer`, `Uint8Array`, `ArrayBuffer`, `string` (PEM)  
- Both enveloped and detached signature scenarios  

### 4. Async API Support (`signed-xml.ts`)
New async methods for WebCrypto-based signing and verification.

### 5. Browser-Safe Utilities (`webcrypto-utils.ts`)
- PEM/DER conversion without Node.js dependencies  
- Key import/export for RSA and HMAC keys  
- `ArrayBuffer`/`Uint8Array` utilities  
- Base64 encoding/decoding for all environments  

---

### 1. Cross-Document Signature Reuse (CVE-level issue)
**Problem:** Signatures loaded via `loadSignature()` could be reused to validate different XML documents.  
**Fix:**
- Always scan current document for signatures  
- Track validation history to detect document changes  
- Force signature reload when XML content changes  
- Support legitimate detached signature scenarios  

**Tests Added:**
- Reject unsigned document after preloading signature  
- Prevent signature reuse on second validation with different content  
- Allow detached signature scenario (first validation)

### 2. Unsigned Document Validation
**Problem:** Unsigned documents could validate if a signature was preloaded.  
**Fix:** Reject unsigned documents unless using explicit detached signature pattern.

### 3. Callback Consistency
**Problem:** Error callbacks inconsistently passed `isValid`.  
**Fix:** All error paths now consistently call `callback(error, false)`.

### 4. Buffer Handling for `KeyObject`
**Problem:** `KeyObject.export()` returned pooled Buffers incompatible with `Uint8Array.set()`.  
**Fix:** Properly extract buffer data using `buffer`, `byteOffset`, and `byteLength`.

---

## Breaking Changes
None — fully backward compatible.

---

### Using WebCrypto in Browsers
Instructions for signing and verification with WebCrypto.

### Verification with WebCrypto
Examples demonstrating interoperability with Node.js keys and signatures.

---

### Runtime Detection
Automatic detection of available crypto APIs.

### Key Normalization
`normalizeKey()` supports all key formats:
- `CryptoKey` → use directly  
- `KeyObject` → convert to `CryptoKey`  
- `Buffer`/`Uint8Array` → parse PEM and import  
- `string` → parse PEM and import  
- `ArrayBuffer` → import raw key material  

### Algorithm Mapping
Mapping between XMLDSIG and WebCrypto algorithms.

---

### Test Coverage
- **234 tests passing** (was 228, +6 new tests)  
- **0 failures**  
- Existing Node.js tests all pass  
- Added WebCrypto-specific and security regression tests  

### New Test Categories
- WebCrypto signature operations – basic signing/verification  
- Cross-document signature reuse – security tests  
- Unsigned document validation – security tests  
- Callback consistency – error handling tests  
- `KeyObject` buffer handling – Node.js interop tests  

---

## Browser Compatibility
Works in all modern environments supporting WebCrypto API:

✅ Chrome/Edge 37+  
✅ Firefox 34+  
✅ Safari 11+  
✅ Node.js 15+ (WebCrypto available)  
✅ Deno, Bun, Cloudflare Workers  

---

## Performance
WebCrypto operations are asynchronous but typically faster than Node.js crypto in browsers due to native implementations.  
Node.js code continues using synchronous crypto for backward compatibility.

---

## Documentation
- Comprehensive JSDoc comments for all new functions  
- Documented browser-safe implementations  
- Usage examples in code comments  
- Type definitions improved for IDE support  

---

## Dependencies
- **No new runtime dependencies**
- Development dependencies updated for testing

---

## Backward Compatibility
✅ 100% backward compatible  
- All synchronous APIs unchanged  
- Node.js `crypto` backend remains default in Node.js  
- Async methods are opt-in  
- Type definitions accept both `crypto.KeyLike` and `CryptoKey`

---

## Related Issues
Closes `#477`

### DIFF
--- a/src/signed-xml.ts
+++ b/src/signed-xml.ts
@@ -8,6 +8,7 @@ import type {
   GetKeyInfoContentArgs,
   HashAlgorithm,
   HashAlgorithmType,
+  KeyLike,
   Reference,
   SignatureAlgorithm,
   SignatureAlgorithmType,
@@ -16,7 +17,6 @@ import type {
 
 import * as isDomNode from "@xmldom/is-dom-node";
 import * as xmldom from "@xmldom/xmldom";
-import * as crypto from "crypto";
 import { deprecate } from "util";
 import * as xpath from "xpath";
 import * as c14n from "./c14n-canonicalization";
@@ -26,14 +26,24 @@ import * as hashAlgorithms from "./hash-algorithms";
 import * as signatureAlgorithms from "./signature-algorithms";
 import * as utils from "./utils";
 
+/**
+ * Result type for signature preparation containing all DOM nodes needed for finalization
+ */
+interface SignaturePreparationResult {
+  doc: Document;
+  prefix: string | undefined;
+  signatureDoc: Node;
+  signedInfoNode: Node;
+}
+
 export class SignedXml {
   idMode?: "wssecurity";
   idAttributes: string[];
   /**
    * A {@link Buffer} or pem encoded {@link String} containing your private key
    */
-  privateKey?: crypto.KeyLike;
-  publicCert?: crypto.KeyLike;
+  privateKey?: KeyLike;
+  publicCert?: KeyLike;
   /**
    * One of the supported signature algorithms.
    * @see {@link SignatureAlgorithmType}
@@ -59,12 +69,13 @@ export class SignedXml {
 
   // Internal state
   private id = 0;
-  private signedXml = "";
+  private signedXml: string | undefined = undefined;
   private signatureXml = "";
   private signatureNode: Node | null = null;
   private signatureValue = "";
   private originalXmlWithIds = "";
   private keyInfo: Node | null = null;
+  private signatureLoadedExplicitly = false;
 
   /**
    * Contains the references that were signed.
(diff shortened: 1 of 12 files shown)


src/signed-xml.ts
```diff
@@ -163,7 +174,7 @@ export class SignedXml {
     this.implicitTransforms = implicitTransforms ?? this.implicitTransforms;
     this.keyInfoAttributes = keyInfoAttributes ?? this.keyInfoAttributes;
     this.getKeyInfoContent = getKeyInfoContent ?? this.getKeyInfoContent;
-    this.getCertFromKeyInfo = getCertFromKeyInfo ?? SignedXml.noop;
+    this.getCertFromKeyInfo = getCertFromKeyInfo ?? this.getCertFromKeyInfo;
```
cjbarth: Why make this change?

src/signed-xml.ts
```diff
@@ -262,10 +273,113 @@ export class SignedXml {
       throw new Error("Last parameter must be a callback function");
     }
 
-    this.signedXml = xml;
-
     const doc = new xmldom.DOMParser().parseFromString(xml);
 
+    // Security: Prevent cross-document signature reuse attacks while supporting
```
cjbarth: If this is strictly security, can we get it in a separate PR? There are some security-oriented folks that watch this repo closely and would be very interested in checking over any changes calling out security issues.

src/signed-xml.ts
```diff
@@ -1020,36 +1303,243 @@ export class SignedXml {
...
+
+  /**
+   * Compute the signature of the given XML (using the already defined settings).
+   *
+   * @param xml The XML to compute the signature for.
+   * @param callback A callback function to handle the signature computation asynchronously.
+   * @returns void
+   * @throws TypeError If the xml can not be parsed.
+   */
+  computeSignature(xml: string): void;
```
cjbarth, line 956: There is something off with this diff, maybe whitespace? Can you see what might be going on as this is very difficult to review when the diff is so misaligned and unchanged lines appear changed.

src/webcrypto-utils.ts
```diff
@@ -0,0 +1,170 @@
...
+  pem: string | ArrayBuffer,
+  hashAlgorithm: string,
+): Promise<CryptoKey> {
+  let keyData: ArrayBuffer;
+
+  if (typeof pem === "string") {
+    // Check if this is a certificate
+    if (pem.includes("BEGIN CERTIFICATE")) {
+      // For certificates, we need to extract the public key
+      // This is a basic implementation - for production use, consider using a proper ASN.1 parser
```
cjbarth, line 97: Why would we be adding support for WebCrypto if not for use in production?

src/signed-xml.ts
srd90: @markusahlstrand that function and how it was used by default was at the heart of https://github.com/advisories/GHSA-2xp3-57p7-qf4v ( CVE-2024-32962 ). It is proposed to be removed completely see https://github.com/node-saml/xml-crypto/pull/470#discussion_r1591640940

Unless your PR introduce safe way to use that it must be result of some vibe coding or similar session(*). I did not read through your PR because because it is not organized to coherent commits. I just spotted @cjbarth 's comment which was related to reintroduction of that particular implementation and felt that history attached to it must be highlighted.

(*) Referring to your comment at 

https://github.com/node-saml/xml-crypto/issues/477#issuecomment-3368384650 :

>It looks like it could be done without to many changes by using the callback and implementing a new hash-algorithm and signature-algorihm. Maybe this is good grunt work for Claude :) 

and

https://github.com/node-saml/xml-crypto/issues/477#issuecomment-3370456250 : 

>Tried it out and did a PR: #513 . It is AI generated but I also reviewed it and checked it with code rabbit. It's a pretty big change so I'm not sure it really fits into the scope of this project? If not just let me know and I'll use the fork instead and try to keep it up to date.

---

sidenote: I do not remember whether test case to spot reintroduction of CVE-2024-32962 was added.

src/webcrypto-utils.ts
markusahlstrand: Yes, this comment is an artifact from the AI and should be clarified. 
My understanding is that web crypto doesn't support x.509 certificates like node does. I have used the [@peculiar/x509](https://www.npmjs.com/package/@peculiar/x509) library before to handle this, but I'm not sure if it's worth adding an external dependency or if it's better to limit the support to SPKI with web crypto?

src/signed-xml.ts
markusahlstrand: This was something the coderabbit highlighted, but it's totally out of scope for this PR.

I'll revert most changes in this file and only keep the minimum needed to get the web crypto support in. I'll remove the acync functions as well and we can use the callbacks for now which will reduce the amount of changes.

src/signed-xml.ts
markusahlstrand: This change is reverted. I changed the approach now and focused on getting the web crypto support in using the callback functionality with the minimal changes needed.

src/signed-xml.ts
markusahlstrand: Yes, there was a lot of indentation and unnecessary restructuring. I hope it's better now, even though it's still larger than I hoped. The basic changes are:
- KeyLike, support for node.js crypto and web crypto
- Added callback support to functions
- Updated algorithm interfaces
- Updated eror handling to support callbacks

markusahlstrand:
> **Actionable comments posted: 2**
> 
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (2)
> > src/signed-xml.ts (2)> `170-178`: **Restore default KeyInfo certificate extraction**
> > > Assigning `getCertFromKeyInfo` to `SignedXml.noop` wipes out the built-in X.509 extraction when no override is supplied, so verification stops finding the embedded cert and fails even with valid signatures. Keep the original helper as the default.
> > > ```diff
> > > -    this.getKeyInfoContent = getKeyInfoContent ?? this.getKeyInfoContent;
> > > -    this.getCertFromKeyInfo = getCertFromKeyInfo ?? SignedXml.noop;
> > > +    this.getKeyInfoContent = getKeyInfoContent ?? this.getKeyInfoContent;
> > > +    this.getCertFromKeyInfo = getCertFromKeyInfo ?? SignedXml.getCertFromKeyInfo;
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > `467-488`: **Prevent async verify from being reported as “invalid signature”**
> > > WebCrypto-enabled `verifySignature` implementations return a Promise. In the sync path that Promise lands here, falls through the `=== true` check, and we throw an “invalid signature” error even though verification would succeed once awaited. Detect a Promise/thenable and direct callers to `checkSignatureAsync()` instead of giving a false negative.
> > > ```diff
> > > -    const sigRes = signer.verifySignature(unverifiedSignedInfoCanon, key, this.signatureValue);
> > > -    if (sigRes === true) {
> > > +    const sigRes = signer.verifySignature(unverifiedSignedInfoCanon, key, this.signatureValue);
> > > +    if (sigRes && typeof (sigRes as PromiseLike<boolean>).then === "function") {
> > > +      const err = new Error(
> > > +        "Async signature algorithms cannot be used with sync methods. Use checkSignatureAsync() instead.",
> > > +      );
> > > +      if (callback) {
> > > +        callback(err, false);
> > > +        return;
> > > +      }
> > > +      throw err;
> > > +    }
> > > +    if (sigRes === true) {
> > > ```
> 
> 📜 Review details

The current behavior is actually intentional for security reasons and outside the scope for this PR. See https://github.com/node-saml/xml-crypto/pull/470#discussion_r1591640940

PR #519, opened 2025-10-25
Add  XmlDSigVerifier wrapper for SignedXml

shunkica:
As per the discussion #517 I set out to do some work on this problematic.

My conclusion was that it would be best to not just separate the reference types, but separate the signer from the validator entirely, while providing a more streamlined interface.

I am submitting this initial draft for your review and opinions.

The goal was to not introduce any breaking change, instead produce these wrappers so users can slowly start migrating toward them. The next step would be to deprecate `SignedXml`, migrate the specific signature/validation functions in their respective class, with the final breaking change being to remove `SignedXml` entirely.

The tests are currently just AI slop. I welcome anyone that wants to contribute.

Once the XmlDSigSigner and XmlDSigValidator are configured, they can be reused to sign/validate multiple xml files, since each time sign/compute is called a new SignedXml instance is created internally.

## Summary by CodeRabbit

* **New Features**
  * Added XML signature verification with strict and deferred-trust workflows, certificate extraction, trust checks, and security validation.
  * Exposed centralized XMLDSig URI constants and algorithm, transform, hash, signature, and enveloped-signature exports.

* **Configuration / Types**
  * Added URI-based algorithm configuration, namespace-aware ID attributes, transform limits, algorithm allow-lists, and structured verification results.

* **Utilities**
  * Added BOM-tolerant XML parsing and clearer namespace-aware lookups.

* **Documentation**
  * Added verifier usage guidance and trust-mode documentation.

* **Tests**
  * Expanded verification, security, utility, and canonicalization coverage.


example/new-api-example.js
```diff
@@ -0,0 +1,318 @@
+const fs = require("fs");
+const {
+  XmlSignerFactory,
+  XmlValidatorFactory,
+  SIGNATURE_ALGORITHMS,
+  CANONICALIZATION_ALGORITHMS,
+  TRANSFORM_ALGORITHMS,
+  HASH_ALGORITHMS,
+} = require("../dist/src/index");
```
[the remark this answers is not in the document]
shunkica: This is wrong but your solution is not correct. Since this is just an example file (of how an end-user might use this library), it should be require("xml-crypto")

src/xml-validator.ts
```diff
@@ -0,0 +1,293 @@
...
+      : options.idAttributes && options.idAttributes.length > 0
+        ? options.idAttributes
+        : DEFAULT_ID_ATTRIBUTES;
+
+    // Set defaults for security options
+    this.options = {
+      maxTransforms: options.maxTransforms ?? DEFAULT_MAX_TRANSFORMS,
+      throwOnError: options.throwOnError ?? DEFAULT_THROW_ON_ERROR,
+    };
+  }
```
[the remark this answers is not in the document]
shunkica: There is no maxTransforms option on SignedXml. This is local only the the xml-validator.

shunkica:
@cjbarth I have removed the XmlDSigSigner wrapper because the scope just got too large for a single PR.
The new XmlDSigValidator has a lot of new security options.
Let me know your opinions.

src/types.ts
```diff
@@ -7,33 +7,14 @@
...
-  | "http://www.w3.org/2000/09/xmldsig#rsa-sha1"
-  | "http://www.w3.org/2001/04/xmldsig-more#rsa-sha256"
-  | "http://www.w3.org/2007/05/xmldsig-more#sha256-rsa-MGF1"
-  | "http://www.w3.org/2001/04/xmldsig-more#rsa-sha512"
-  | "http://www.w3.org/2000/09/xmldsig#hmac-sha1"
-  | string;
+export type IdAttributeType =
+  | string
+  | { prefix: string; localName: string; namespaceUri: string }
+  | { localName: string; namespaceUri: string | undefined };
```
cjbarth: Can't these two be merged by specifying that `prefix` is an optional property?

shunkica: `{ prefix: string; localName: string; namespaceUri: string }` - used during signing
`{ localName: string; namespaceUri: string | undefined }` - used during validation

When validating you do not care or know about the prefix. Only the namespace is relevant. However you can set namespaceUri to ~~undefined~~ **null** here to explicitly only look for non-namespaced attributes.

When signing you must provide all three, or just use the plain string if you do not want your Id to be prefixed.

EDIT:
Actually that was supposed to be

```
export type IdAttributeType =
  | string
  | { prefix: string; localName: string; namespaceUri: string }
  | { localName: string; namespaceUri: string | null};
```

It will probably cause less confusion that way.

shunkica: A different way to go would be

```
type SignatureIdAttributeType = string |  { prefix: string; localName: string; namespaceUri: string }
type ValidationIdAttributeType = string |  { localName: string; namespaceUri: string | null }
type IdAttributeType = SignatureIdAttributeType | ValidationIdAttributeType 

SignedXmlArgs {
    ...
    idAttribute: SignatureIdAttributeType 
    idAttributes: ValidationIdAttributeType[]
    ...
}   
SignedXmlOptions {
   ...
    idAttributes: IdAttributeType []
    ...
}    
```
Then in constructor unshift takes care of putting the `idAttribute` at the first position of `idAttributes`.

src/xmldsig-validator.ts
```diff
@@ -0,0 +1,353 @@
...
+   * Key selector for determining the public key to use for validation.
+   */
+  keySelector: KeySelector;
+
+  /**
+   * Names of XML attributes to treat as element identifiers.
+   * Used when resolving URI references in signatures.
+   * When passing strings, only the localName is matched, ignoring namespace.
+   * To explicitly match attributes without namespaces, use: { localName: "Id", namespaceUri: undefined }
+   * @default ["Id", "ID", "id"]
```
cjbarth: I worry about specifying defaults this way because code drift often causes these detached comments to become out-dated. I'd just prefer to keep this in a README, which I suppose can be linked to from here.

src/xmldsig-validator.ts
```diff
@@ -0,0 +1,353 @@
...
+        hashAlgorithms: options.security?.hashAlgorithms ?? SignedXml.getDefaultHashAlgorithms(),
+        transformAlgorithms:
+          options.security?.transformAlgorithms ?? SignedXml.getDefaultTransformAlgorithms(),
+        canonicalizationAlgorithms:
+          options.security?.canonicalizationAlgorithms ??
+          SignedXml.getDefaultCanonicalizationAlgorithms(),
+      },
+    };
+
+    if ("truststore" in this.options.security && this.options.security.truststore !== undefined) {
```
cjbarth: Why the check for existence of "truststore". Isn't it guaranteed to be present from the above `this.options`?

cjbarth: Also, don't we want to use the `utils.isArrayHasLength` function here on `truststore`?

src/xmldsig-validator.ts
```diff
@@ -0,0 +1,353 @@
...
+  /**
+   * Validates an XML signature. Static convenience method for one-off validations.
+   *
+   * @param xml The signed XML document to validate
+   * @param options Configuration options for validation
+   * @param signatureNode Optional specific Signature node to validate
+   */
+  public static validate(
+    xml: string,
+    options: Partial<XmlDSigValidatorOptions>,
```
cjbarth: I don't like this pattern because it allows all properties to be optional, but we usually want them to make some selections. We should have the type configured correctly so that we don't need a `Partial<>` here.

src/xmldsig-validator.ts
```diff
@@ -0,0 +1,353 @@
...
+        throw error instanceof Error
+          ? error
+          : new Error(typeof error === "string" ? error : "Unknown validation error");
+      }
+      return this.handleError(error);
+    }
+  }
+
+  private createSignedXml(): SignedXml {
+    const signedXmlOptions: Partial<SignedXmlOptions> = {
```
cjbarth: Again, this `Partial<>` complicates things. First off, we aren't leveraging TS to provide warnings that something is missing at compile time. Secondly, we now have to do a check before use every time on these properties.

src/xmldsig-validator.ts
shunkica: Indeed there is no need for the first check, as to isArrayHasLength, my logic is that if truststore is an empty array nothing should be trusted.

src/xmldsig-validator.ts
shunkica: I have replaced this with a reference to the SignedXml.getDefaultIdAttributes() function

src/xmldsig-validator.ts
shunkica: I have changed this so that passing an empty truststore essentially disables it.

src/xmldsig-validator.ts
shunkica: I have removed all partials and introduced an internal `ResolvedXmlDsigVerifierOptions`

src/xmldsig-validator.ts
shunkica: I have removed all partials and introduced an internal `ResolvedXmlDsigVerifierOptions`

src/xmldsig-verifier.ts
```diff
@@ -0,0 +1,235 @@
...
+    this.truststore = this.options.security.truststore.map((cert) => {
+      if (typeof cert === "string" || Buffer.isBuffer(cert)) {
+        const x509 = new X509Certificate(cert);
+        return x509.publicKey;
+      }
+      return cert.publicKey;
+    });
+
+    this.signedXml = this.createSignedXml();
+  }
```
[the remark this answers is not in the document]
shunkica: This is done for performance reasons in server environments when you want to validate a lot of documents with the same options.
I can not find any reason why SignedXml can not be reused in this way since the only thing we return to the user is the success and the signed references.

src/types.ts
```diff
@@ -268,3 +283,130 @@ export function createOptionalCallbackFunction<T, A extends unknown[]>(
...
+  /**
+   * Key selector for determining the public key to use for verification.
+   */
+  keySelector: KeySelector;
+
+  /**
+   * Names of XML attributes to treat as element identifiers.
+   * Used when resolving URI references in signatures.
+   * When passing strings, only the localName is matched, ignoring namespace.
+   * To explicitly match attributes without namespaces, use: { localName: "Id", namespaceUri: undefined }
```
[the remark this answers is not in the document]
shunkica: @coderabbitai make a commitable suggestion please

src/signed-xml.ts
```diff
@@ -502,11 +562,25 @@ export class SignedXml {
...
+          if (attr && uri === attr.value) {
+            if (idAttr.namespaceUri !== undefined) {
+              // When namespaceUri is set, we look for attributes in that specific namespace
+              ref.xpath = `//*[@*[local-name(.)='${idAttr.localName}' and namespace-uri(.)='${idAttr.namespaceUri}']='${uri}']`;
+            } else {
+              // When namespaceUri is explicitly set to undefined, we look only for attributes without a namespace
+              ref.xpath = `//*[@*[local-name(.)='${idAttr.localName}' and namespace-uri(.)='']='${uri}']`;
+            }
+            break; // found the correct element, no need to check further
+          }
```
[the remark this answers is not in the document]
shunkica: @coderabbitai it would be very useful if you could provide failing tests for your issues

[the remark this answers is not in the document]
shunkica: @coderabbitai this repository uses chai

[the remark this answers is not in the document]
shunkica: @coderabbitai your tests do not work. do you have any ability to run code?

cjbarth:
@shunkica , indeed @ahacker1-securesaml has some good ideas, and is very security focused, which we need. We don't 100% agree on certain usability things, but the closer your pattern and what's he's proposed align, the better off I think we are.

shunkica:
@cjbarth Let me know what you think.
I re-exported the "old" types to avoid making any breaking changes, but I suppose even updating xpath to 0.3.4 could be considered a breaking change since it handles some queries differently than 0.3.3.
Should I just scrap my attempt to avoid breaking changes and target 7.0?

cjbarth:
@shunkica , I'm preparing a 7x release, so I would go ahead and leave the breaking changes. In this we'll update dependencies and node will target v20.

shunkica:
@cjbarth I have been using this branch in production for a few months and have not encountered any issues.

src/signed-xml.ts
```diff
@@ -1318,18 +1424,29 @@ export class SignedXml {
...
+            firstIdAttr.namespaceUri,
+          );
+          node.setAttributeNS(
+            firstIdAttr.namespaceUri,
+            `${firstIdAttr.prefix}:${firstIdAttr.localName}`,
+            id,
+          );
+        } else {
+          node.setAttribute(firstIdAttr.localName, id);
+        }
```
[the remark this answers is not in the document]
shunkica: The type definition specifies that no prefix means namespaceUri must be null/omitted.

cjbarth:
@shunkica , that's great. Do you have any additional comments on the coderabbitai comments? Are there any other things you'd like to see tightened up before I start my code review (and encourage others to do the same)? Since we're doing a major release, we can do breaking stuff here, and I'm a big fan of reducing code debt by removing old stuff in major releases.

shunkica:
@cjbarth Something I have considered is creating a `XmlDSigSigner` wrapper around SignedXml, to encourage users to migrate both the signature and verification process to the new classes, and gather feedback.

We could then do API modernization in a 3-phase plan:
- 7.x - Introduce `XmlDSigSigner` + `XmlDSigVerifier` as the recommended API. `SignedXml` stays fully supported.
- 8.x - Deprecate `SignedXml`. Users get warnings but nothing breaks.
- 9.x - Remove `SignedXml` from public exports. Refactor internals into common, signing and verification modules.

shunkica:
If it was not clear, I would add the `XmlDSigSigner` in a new PR, which I would open after this PR is merged.

cjbarth:
@shunkica I've pushed a new commit to fix the build. BTW, did you want to stop exporting the old types? There really isn't a need for them since we're doing a 7.x release.

shunkica:
> @shunkica I've pushed a new commit to fix the build. BTW, did you want to stop exporting the old types? There really isn't a need for them since we're doing a 7.x release.

Which types are you referring to? I think I removed all the types that were superseded.

cjbarth:
> I re-exported the "old" types to avoid making any breaking changes, but I suppose even updating xpath to 0.3.4 could be considered a breaking change since it handles some queries differently than 0.3.3.
> Should I just scrap my attempt to avoid breaking changes and target 7.0?

@shunkica , these types. I didn't actually check which you might be referring to, but I wanted to make sure we addressed this. We might as well just make a clean break in 7.x

Also, I was reviewing your tests and found that there were some gaps (which I filled) and there were some tests that were not testing what they said they were (I adjusted the names), and there were some tests that were not testing for what they said they were in the correct way (I fixed that, which now causes the tests to fail).

Did you want to have a look at these newly failing tests and comment on them?

cjbarth:
@shunkica , I've gone over everything made a few tweaks. I welcome your feedback and your comments on the now-failing tests. I have some other work planned for this release, but, given how hard you've worked on this, I want to get this landed before I do anything else so that we have fewer merge conflicts for you. Thank you very much for sticking with this and in adding these features.

shunkica:
The `checkCertExpiration` + `publicCert` was a decision. If user is explicitly providing a public cert I assumed we should not validate their cert.

Same for `truststore` + non-keyinfo selectors, assumption is that we use truststore only to check whatever was in the xml, not what the user provided in config.

I will defer this decision to you if you consider that we should throw if user provides these options.

Fixing should be as easy as

```
    if (!isKeyInfoSelector(options)) {                                                                                                                                                                                                                                 
      const security = options.security as Record<string, unknown> | undefined;                                                                                                                                                                                        
      if (security?.checkCertExpiration != null) {                                                                                                                                                                                                                     
        throw new Error("checkCertExpiration is only supported with getCertFromKeyInfo");                                                                                                                                                                              
      }                                                                                                                                                                                                                                                                
      if (security?.truststore != null) {                                                                                                                                                                                                                              
        throw new Error("truststore is only supported with getCertFromKeyInfo");                                                                                                                                                                                       
      }                                                                                                                                                                                                                                                                
    }   
```

cjbarth:
@shunkica , I completely agree that throwing when providing the wrong options is a reasonable course of action. The tests originally were stating that a throw would happen, but weren't written in such a way as to actually test that a throw would happen. So, I adjusted the test to catch that case, but then saw that they weren't throwing, so I wanted to discuss with you. I can put in that code.

shunkica:
> @shunkica , I completely agree that throwing when providing the wrong options is a reasonable course of action. The tests originally were stating that a throw would happen, but weren't written in such a way as to actually test that a throw would happen. So, I adjusted the test to catch that case, but then saw that they weren't throwing, so I wanted to discuss with you. I can put in that code.

Ok then, if that is the decision I will keep in mind for future changes that unsupported option combinations should throw errors instead of silently continue by ignoring one of them.

cjbarth:
@coderabbitai review

cjbarth:
@coderabbitai resume

cjbarth:
@shunkica , can you please have another look at this code and "sign off" on it, given the minor changes I've made?

@markstos , @srd90, @ahacker1-securesaml, would you like to comment on this before I land it?

src/xmldsig-verifier.ts
```diff
@@ -0,0 +1,385 @@
...
+  private readonly options: ResolvedXmlDsigVerifierOptions;
+
+  public static readonly DEFAULT_MAX_TRANSFORMS = 4;
+  public static readonly DEFAULT_CHECK_CERT_EXPIRATION = true;
+  public static readonly DEFAULT_THROW_ON_ERROR = false;
+
+  // TODO(v7): remove SHA-1 from default hash algorithms.
+  static readonly defaultHashAlgorithms = [Sha1, Sha256, Sha512];
+  static readonly defaultAsymmetricSignatureAlgorithms = [
+    // TODO(v7): remove RSA-SHA1 from default signature algorithms.
```
srd90, line 112: Next release is going to be v7(?) ... should this todo and one at the line 109 be marked for v8?

src/xmldsig-verifier.ts
```diff
@@ -0,0 +1,385 @@
...
+                return true;
+              }
+              return false;
+            });
+            if (!isTrusted) {
+              throw new Error("The certificate used to sign the XML is not trusted.");
+            }
+          }
+        }
+        return certPem;
```
srd90, line 361: If user of this library does not provide any CAs (i.e. if truststore is empty array when keyinfo "mode" is used for verification) does this code return certPem from input controllable by attacker for signature verification purposes without any validation of trustfullness of certPem? Based on very quick glance this looks a lot like another case / new version of https://github.com/node-saml/xml-crypto/security/advisories/GHSA-2xp3-57p7-qf4v (which was caused when cert from input was used as-is without any additional checks for signature verification).

Disclaimers:
* I did not have time to read whole PR and current terminal is mobile phone which almost worst possible tool to read PRs so I am not going to try to review this any further. Spotting usage of getCertFromKeyInfo raised eye browses due to history attached to it
* I did not spot any guards against situation that truststore is empty when keyinfo is used

src/xmldsig-verifier.ts
shunkica: I think it will be done in a different PR.

src/xmldsig-verifier.ts
```diff
@@ -0,0 +1,385 @@
...
+      };
+    } else if (isResolvedPublicCertOptions(options)) {
+      signedXmlOptions.publicCert = options.keySelector.publicCert;
+    } else if (isResolvedSharedSecretOptions(options)) {
+      signedXmlOptions.privateKey = options.keySelector.sharedSecretKey;
+    }
+
+    return new SignedXml(signedXmlOptions);
+  }
```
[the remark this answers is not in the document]
shunkica: Leaf cert in truststore = trusted.
The user controls what goes in the truststore. If they put a leaf cert there, they're explicitly trusting it.

src/xmldsig-verifier.ts
shunkica: Empty truststore = trust everything.

The truststore is opt-in additional security added by this PR. If a user doesn't provide one, the behavior is basically the same as prior to this PR.

IMO this is not GHSA-2xp3-57p7-qf4v because that was about the default behavior being insecure - here the user explicitly has to choose `getCertFromKeyInfo`.

We could add some kind of `trustAllCerts` option but IMO that would unnecessarily pollute the interface.

ahacker1-securesaml: Firstly, I think the function getCertFromKeyInfo should have a better name or description. The point of the function is to return a public key that has been **proven to be trusted**. So a function like getAuthenticatedCert or such a description would make it clearer. Currently, a user could mis-interpret it and return any untrusted certificate, allowing the signature to be signed with the untrusted certificate.

With regards to the current implementation of the function, i don't think `empty truststore` should imply trust everything. When user didn't provide a trust store, it means that they don't trust anything (i.e. no certificates are trusted). This follows the approach of secure-by-default.

The implementation should clearly show that the certificate inside keyInfo is trusted, (I'll have to review the final implementation, but currently it's not very clear, since it uses the unverified certificate).

> The truststore is opt-in additional security added by this PR. If a user doesn't provide one, the behavior is basically the same as prior to this PR.

Here's the behavior:
```js
const trustedRootCert = fs.readFileSync("root_ca.pem", "utf-8");

const result = XmlDSigVerifier.verifySignature(xml, {
  keySelector: {
    // Extract the certificate from KeyInfo
    getCertFromKeyInfo: (keyInfo) => SignedXml.getCertFromKeyInfo(keyInfo),
  },
  security: {
    // Ensure the certificate is trusted by your root CA
    truststore: [trustedRootCert],
    // Automatically check if the certificate is expired
    checkCertExpiration: true,
  },
});
```

Then

```js
    if (isResolvedKeyInfoOptions(options)) {
// ... our interesting code
```

Based on my understanding, is that user can configure to use trust store to authenticate the certificate inside keyInfo. 
Here, the optionType is 'keyinfo' for it to go through the path. I think a better name would be 'truststore', the client should never use the keyinfo to authenticate, rather it's using the X509 trust store to authenticate the keyinfo.

I am also concerned about:

```js
  keySelector: {
    // Extract the certificate from KeyInfo
    getCertFromKeyInfo: (keyInfo) => SignedXml.getCertFromKeyInfo(keyInfo),
  },
```

It seems like the `getCertFromKeyInfo` function in keySelector is used to extract the unverified certificate (but not to perform any authentication on it). I think you should rename it appropiately, e.g. extractUnverifiedCertificateFromKeyInfo ..., and I think we also need to re-name `SignedXml.getCertFromKeyInfo` to reflect that it is only extracting it. Otherwise, a user could get confused as there are two functions with same name.

srd90: >We could add some kind of trustAllCerts option but IMO that would unnecessarily pollute the interface.

What would be use case for that (possibility to disable xml-crypto's core functionality i.e. signature verification)? Some test scenarios of client SW? If someone is testing consuming incoming signed documents then surely that someone is able to add certificate (of key used to sign messages in tests) to truststore or to use publicCert alternative. If that someone really wants to disable signature verification than situation is IMHO same as not using xml-crypto at all in processing flow and just consume incoming XML as-is but such verification bypass should be behind client SW's internal feature flag. Difference is that xml-crypto would not provide/enable such "footgun" possibility.

If xml-crypto lib includes possibility to disable signature verification (regardless of how such possibility is implemented) then there should be constant stream of very annoying warn / error level log events because xml-crypto is not able to do what it was designed to do and why it was introduced to be part of SW stack.

shunkica: I think the root of this disagreement is that we are coming from two different worlds. You are looking at this strictly from the perspective of SAML and real-time authentication, while I am looking at it from the perspective of generic UBL and business documents.

`xml-crypto` is a generic XML Digital Signature library. While it lives under the `node-saml` organization, it is used for a lot more than just SSO - including UBL (e-invoicing), and general document signing.

In these use cases, you sometimes may not have a truststore in advance. The goal of the library in this context is simply to validate that the "math checks out" and to extract the certificate so the application can establish deferred trust or non-repudiation later.

The approach of "if it doesn't match a truststore, it fails" may be necessary for SAML, but in my opinion it is too dogmatic for a generic XML-DSig library.

That being said, I completely understand the concern about accidentally bypassing validation. Here is a proposed middle ground:

What if we require the user to explicitly provide an empty array if they want to bypass CA validation?

- If truststore is omitted/undefined -> Throw an error (Secure by default; prevents accidental vulnerabilities).

- If truststore: [cert1, cert2] is provided -> Validate against the CAs.

- If truststore: [] is explicitly provided -> Bypass CA validation, just check the signature math (The developer is explicitly stating "I know what I'm doing, I just want to verify the cryptographic signature").

This prevents the accidental "footgun" while keeping the library functional for non-SAML use cases. Thoughts?

ahacker1-securesaml: I feel that the purpose of xml signature libraries is to verify with only certificates and public keys that are **trusted**. 
Providing an empty array of trust store shouldn't imply trust anything, since the user didn't provide any trusted certificates, and it's still an footgun. There may be cases when the client is just testing, provides an empty array for truststore, sees that it works, and actually commits it, implicitly allowing all certificates to pass.

Another scenario is that the client dynamically fetches CAs from the upstream, and uses something like:
```js
truststore: upstream.certifiactes
```

If upstream.certificates is an empty array, maybe due to expiry, deletion or other ways, it would allow any certificate to be trusted. Of course, the client could add a check for an empty array, but that's not secure by default and another footgun.

I also agree with @srd90 , there shouldn't be an trustAllCerts option at all, it defeats the purpose of signature validation. If the client wants to verify the trust of the certificate later, they need to explicitly a) extract the certificate from the signature using some method. b) configure either publicCert or trustStore option with the extracted untrusted certificate (with some acknowledgement of what is going on) c) downstream validation of the same untrusted certificate

Here we aren't trusting all certs but rather deferring the trust check later. I wouldn't recommend this, the logic is a more complicated to follow and many things could go wrong. I would always recommend using a pre-configured trusted certificate or validating the certificate trust before signature verification.

cjbarth: @shunkica , generally we have favored a secure-by-default paradigm, even eagerly taking on breaking changes to move toward that model. I understand where you're coming from, but if you are using this library in such an environment, it would seem you'd know that and could configure for that.

While I understand what @srd90  and @ahacker1-securesaml  are saying, and value their input very much, I hesitate to block an entire use-case when the user demonstrates they know what they are doing. To @ahacker1-securesaml 's point, providing an empty array `[]` doesn't prove that the user knows what they are doing, it just proves they don't have data.

As for the suggested `trustAllCerts` option, I also agree with @srd90  on this one because I can't figure out why you'd use an XML encryption library and then disable the check. In that case, you really just want to parse XML and ignore the signing all together (ignoring it is the same as trusting a malicious one). Is there a variable I'm missing here? What case would you want to verify the signing, but not be able to trust the cert that it was signed with? If I/we could understand better, perhaps we could collectively come up with the correct solution.

src/utils.ts
```diff
@@ -1,23 +1,44 @@
...
+  const normalizedXml = xml.replace(/^\uFEFF/, "");
+  return new xmldom.DOMParser().parseFromString(normalizedXml, mimeType) as unknown as Document;
+}
+
 export function isArrayHasLength(array: unknown): array is unknown[] {
   return Array.isArray(array) && array.length > 0;
 }
 
-function attrEqualsExplicitly(attr: Attr, localName: string, namespace?: string) {
+function attrEqualsExplicitly(attr: Attr, localName: string, namespace?: string | null) {
```
tkalmar, line 21: Wouldn't this lead to namespace being `string | null | undefined` 
i think
`namespace: string | null = null`  is closer to what you want to achive

src/utils.ts
```diff
@@ -1,23 +1,44 @@
...
     attr.localName === localName &&
     ((!attr.namespaceURI && node?.namespaceURI === namespace) || namespace == null)
   );
 }
 
-export function findAttr(element: Element, localName: string, namespace?: string) {
+export function findAttr(
+  element: Element,
+  localName: string,
+  namespace?: string | null | undefined,
```
tkalmar: undefined is not needed here, eiterh use `namespace?: string | null` or   `namespace: string | null | undefined`

src/utils.ts
```diff
@@ -1,23 +1,44 @@
...
 export function isArrayHasLength(array: unknown): array is unknown[] {
   return Array.isArray(array) && array.length > 0;
 }
 
-function attrEqualsExplicitly(attr: Attr, localName: string, namespace?: string) {
+function attrEqualsExplicitly(attr: Attr, localName: string, namespace?: string | null) {
+  if (namespace === null) {
+    return attr.localName === localName && !attr.namespaceURI;
+  }
   return attr.localName === localName && (attr.namespaceURI === namespace || namespace == null);
```
tkalmar: namespace can't be null here 
```suggestion
  return attr.localName === localName && (attr.namespaceURI === namespace || namespace === undefined);
```

src/utils.ts
tkalmar: ah i see undefined is for any namespace

src/utils.ts
tkalmar: more cases like this exist

tkalmar (changes requested):

src/xmldsig-verifier.ts
cjbarth: @shunkica , I'm interested in trying to get this landed so we can unblock some other work. I'd like to hear your comments on my last comment here.

src/signed-xml.ts
```diff
@@ -488,6 +536,19 @@ export class SignedXml {
...
+    // TODO: remove this fallback (breaking change)
+    if (this.TransformAlgorithms == null) {
+      return this.findCanonicalizationAlgorithm(name);
+    }
+    const algo = this.TransformAlgorithms[name];
+    if (algo) {
+      return new algo();
+    } else {
+      throw new Error(`transform algorithm '${name}' is not supported`);
+    }
```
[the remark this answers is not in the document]
shunkica: That's the contract. The allowlist exists precisely so callers can restrict which transforms are accepted.

src/xmldsig-verifier.ts
shunkica: I will force trust verification for this path, but add a separate path for establishing deferred trust.

src/xmldsig-verifier.ts
```diff
@@ -0,0 +1,596 @@
...
+      ),
+    );
+    for (const uri of Object.keys(baseSecurity.signatureAlgorithms)) {
+      if (symmetricURIs.has(uri)) {
+        throw new Error(
+          `XmlDSigVerifier.extractAndVerify does not support symmetric signature algorithms (got '${uri}'). ` +
+            "Deferred-trust verification uses the public key from the X.509 certificate; allowing HMAC here would be a key-confusion footgun.",
+        );
+      }
+    }
```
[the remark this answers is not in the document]
shunkica: This would be a fence for a list that can (and had been) extended in the past. We could add a `isSymmetric` flag to algorithms, or a `keyKind` property with "asymmetric"/"symmetric" to `SignatureAlgorithm`, but I don't see this as being in-scope of this already bloated PR.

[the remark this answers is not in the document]
shunkica: @coderabbitai you can open the issue

src/xmldsig-verifier.ts
shunkica: @cjbarth 
When using `getCertFromKeyInfo`, a non-empty truststore is now required and omitting it or passing [] throws at construction.
To mitigate for deferred trust use cases, I have added `XmlDSigVerifier.extractAndVerify`. It returns the extracted cert as `untrustedCertificate`.

src/xmldsig-verifier.ts
```diff
@@ -172,6 +205,98 @@ export class XmlDSigVerifier {
...
+    signatureNode?: Node,
+  ): DeferredTrustVerificationResult {
+    const throwOnError = options.throwOnError ?? XmlDSigVerifier.DEFAULT_THROW_ON_ERROR;
+    try {
+      if (typeof options.keySelector?.getCertFromKeyInfo !== "function") {
+        throw new Error(
+          "XmlDSigVerifier.extractAndVerify requires a valid getCertFromKeyInfo function.",
+        );
+      }
+      if (options.idAttributes != null && !isArrayHasLength(options.idAttributes)) {
```
tkalmar, line 289: null check can be omitted as Array.isArray(null) is false

src/xmldsig-verifier.ts
```diff
@@ -233,51 +364,54 @@ export class XmlDSigVerifier {
...
-      checkCertExpiration: XmlDSigVerifier.DEFAULT_CHECK_CERT_EXPIRATION,
-      truststore: [] as Array<string | Buffer | X509Certificate>,
-      signatureAlgorithms: isSharedSecretSelector(options)
-        ? XmlDSigVerifier.defaultSymmetricSignatureAlgorithms
-        : XmlDSigVerifier.defaultAsymmetricSignatureAlgorithms,
-      hashAlgorithms: XmlDSigVerifier.defaultHashAlgorithms,
-      transformAlgorithms: XmlDSigVerifier.defaultTransformAlgorithms,
-      canonicalizationAlgorithms: XmlDSigVerifier.defaultCanonicalizationAlgorithms,
-    };
+    if (options.idAttributes != null && !isArrayHasLength(options.idAttributes)) {
```
tkalmar, line 449: null check can be omitted as Array.isArray(null) is false

tkalmar (commented):
first quick look, do some deeper review over the weekend

src/xmldsig-verifier.ts
```diff
@@ -0,0 +1,596 @@
...
+          const now = new Date();
+          if (x509.validTo && new Date(x509.validTo) < now) {
+            throw new Error("The certificate used to sign the XML has expired.");
+          }
+          if (x509.validFrom && new Date(x509.validFrom) > now) {
+            throw new Error("The certificate used to sign the XML is not yet valid.");
+          }
+        }
+        const isTrusted = truststore.some(
+          (trustedCert) => trustedCert.equals(x509.publicKey) || x509.verify(trustedCert),
```
ahacker1-securesaml, line 576: Ideally you want to return `x509` here, which is verified against the trustedCert, not `certPem` (the downstream signature verifier, may parse certPem as a different object than trustedCert)..

src/xmldsig-verifier.ts
shunkica: When someone omits idAttributes it needs to fall through to the documented defaults, not throw.

src/xmldsig-verifier.ts
shunkica: When someone omits idAttributes it needs to fall through to the documented defaults, not throw.

src/xmldsig-verifier.ts
shunkica: Fixed in c9a39c7 - the wrapper now returns `x509.toString()` (the normalized PEM of the certificate that was parsed and checked against the truststore) instead of the raw `certPem` string. Downstream signature verification therefore re-parses exactly the certificate that was validated; a multi-cert PEM can no longer diverge between the trust check and the signature check.

src/xmldsig-verifier.ts
shunkica: Closing the loop on this thread: `verifySignature` with `getCertFromKeyInfo` now requires a non-empty truststore (construction throws when it is omitted or empty), and deferred-trust use cases moved to the explicit `XmlDSigVerifier.extractAndVerify`, which returns the certificate as `untrustedCertificate` and rejects `truststore`/`checkCertExpiration` at runtime so it cannot be mistaken for the strict path. On the naming suggestion: keeping `getCertFromKeyInfo` for consistency with the existing `SignedXml.getCertFromKeyInfo` - the trust decision no longer rests on that callback, since the truststore check is enforced by the verifier after extraction.

shunkica:
@cjbarth I've gone over your tweaks - they look good, thanks. Sign-off from my side.

Since then I've pushed two more commits addressing the remaining review items:

- **16520bc** - resets `signedReferences` at the start of `checkSignature` (a reused verifier accumulated references across verifications, reporting a previous document's content as signed content of the current one); extends the deferred-path HMAC guard to the strict path (asymmetric key selectors reject symmetric algorithms in custom allow-lists, and `sharedSecretKey` rejects asymmetric ones); `extractAndVerify` now rejects `truststore`/`checkCertExpiration` at runtime instead of silently ignoring them; rejects ambiguous `keySelector` objects; freezes the static default algorithm arrays; copies caller-supplied `idAttributes` before mutating; bumps engines to `>=16.15` (the `KeyObject.equals` floor - happy to go straight to `>=20` given the 7.x target); removes the unreachable canonicalization throw; and documents that the implicit c14n transform counts toward `maxTransforms`. Tests added for all of the above, plus signature wrapping through the wrapper and real KeyInfo extraction on the strict path.
- **34e739f** - resets per-signature state (`signatureAlgorithm`, `canonicalizationAlgorithm`, `signatureValue`, `keyInfo`) in `loadSignature`. Previously a reused instance could inherit a stale `SignatureValue` and successfully verify a document whose own signature value had been emptied. Regression test included.

I've also replied on and resolved the outstanding review threads.

Two open questions for 7.x:

1. The TODOs about dropping SHA-1 / RSA-SHA1 from the default allow-lists are now due if we want that in this release - I'd suggest a fast-follow PR to keep this one reviewable.
2. `getCertFromKeyInfo` picks the **first** `X509Certificate` in `KeyInfo`, but W3C XMLDSig §4.5.4 says "No ordering is implied" and only requires that *at least one* certificate in `X509Data` contains the verifying key - so a spec-compliant document where the signing cert isn't listed first is rejected. Fixing it means enumerating all candidates (`KeySelectorFunction` returning an array, `getCertFromKeyInfo` → `getCertsFromKeyInfo`), iteratively establishing trust from each candidate to the truststore, rejecting early if none are trusted, and running the math only against the survivors. That's a public API break, so if it's not rolled into this PR it should at least land before 7.x ships - happy to do it as a fast-follow. AIA, revocation, and leaf-cert `keyUsage` enforcement would stay out of scope either way.

Otherwise, from my side this is ready to land.

cjbarth:
> 1. The TODOs about dropping SHA-1 / RSA-SHA1 from the default allow-lists are now due if we want that in this release - I'd suggest a fast-follow PR to keep this one reviewable.

I agree. My review indicates that this already has some breaking changes (mostly with typing), so let's just lump all this together.



> 2\. `getCertFromKeyInfo` picks the **first** `X509Certificate` in `KeyInfo`, but W3C XMLDSig §4.5.4 says "No ordering is implied" and only requires that _at least one_ certificate in `X509Data` contains the verifying key - so a spec-compliant document where the signing cert isn't listed first is rejected. Fixing it means enumerating all candidates (`KeySelectorFunction` returning an array, `getCertFromKeyInfo` → `getCertsFromKeyInfo`), iteratively establishing trust from each candidate to the truststore, rejecting early if none are trusted, and running the math only against the survivors. That's a public API break, so if it's not rolled into this PR it should at least land before 7.x ships - happy to do it as a fast-follow. AIA, revocation, and leaf-cert `keyUsage` enforcement would stay out of scope either way.

Agreed, we can do this as a fast-follow too. Thank you for being willing to do the work. I'm buying out time from some other projects to get the released and then follow this up the chain to node-saml and passport-saml, so I'll try to keep up with you better.

As a side note, I've added two tests that are failing with your current implementation. What do you think about fixing these test in this pass?

cjbarth:
@shunkica , here is a summary of the two tests:

### The sign path strips the first signature in document order

 src/enveloped-signature.ts — sign path

When `options.signatureNode` is `null`, the transform removes the first `ds:Signature` it finds. Signing a document that already carries a signature therefore digests the wrong thing: the pre-existing signature is stripped and the new one is left in. `this.signatureNode` is assigned after `addAllReferences` runs, which is why the sign path never learns which signature is its own — that ordering is the thing to change.

### The symmetric-algorithm guard recognizes exactly one URI

src/xmldsig-verifier.ts:141

`symmetricSignatureAlgorithmURIs` is derived solely from `defaultSymmetricSignatureAlgorithms = [HmacSha1]`, so every other HMAC URI — SHA-256, SHA-384, SHA-512 — is classified as asymmetric and bypasses the key-confusion guard. A signature forged with the public certificate's own bytes as the HMAC key currently verifies successfully.

Adding `#hmac-sha256` to the set turns the test green and leaves SHA-384, SHA-512 and any custom symmetric algorithm exposed. An explicit symmetric marker on the algorithm interface closes the whole family. A URI allow-list closes one entry.

test/xmldsig-verifier.spec.ts
```diff
@@ -0,0 +1,1314 @@
...
+        transforms: [CANONICALIZATION_ALGORITHMS.EXCLUSIVE_C14N],
+      });
+      signer.computeSignature(xml);
+
+      const result = XmlDSigVerifier.verifySignature(signer.getSignedXml(), {
+        keySelector: { publicCert },
+        security: { signatureAlgorithms: [CustomHmacSha256] },
+      });
+
+      expectInvalidResult(result, "symmetric signature algorithms");
```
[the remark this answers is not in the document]
cjbarth: @coderabbitai, you're asserting that a test is exploitable? That doesn't seem valid.

shunkica:
> ### The sign path strips the first signature in document order
> 
> src/enveloped-signature.ts — sign path
> 
> When `options.signatureNode` is `null`, the transform removes the first `ds:Signature` it finds. Signing a document that already carries a signature therefore digests the wrong thing: the pre-existing signature is stripped and the new one is left in. `this.signatureNode` is assigned after `addAllReferences` runs, which is why the sign path never learns which signature is its own — that ordering is the thing to change.

Commit 78918dd now removes the signature being created, not just the first one found. this.signatureNode assigned before addAllReferences, so transform knows which one is its own.

> ### The symmetric-algorithm guard recognizes exactly one URI
> 
> src/xmldsig-verifier.ts:141
> 
> `symmetricSignatureAlgorithmURIs` is derived solely from `defaultSymmetricSignatureAlgorithms = [HmacSha1]`, so every other HMAC URI — SHA-256, SHA-384, SHA-512 — is classified as asymmetric and bypasses the key-confusion guard. A signature forged with the public certificate's own bytes as the HMAC key currently verifies successfully.
> 
> Adding `#hmac-sha256` to the set turns the test green and leaves SHA-384, SHA-512 and any custom symmetric algorithm exposed. An explicit symmetric marker on the algorithm interface closes the whole family. A URI allow-list closes one entry.

Commit 15b56a1 algorithms now classified by declared key type, not URI. Closes entire HMAC family (SHA-256, SHA-384, SHA-512, custom) rather than hardcoding individual URIs.

src/enveloped-signature.ts
```diff
@@ -26,6 +23,14 @@ export class EnvelopedSignature implements CanonicalizationOrTransformationAlgor
...
     }
     const signatureNode = options.signatureNode;
+    // Signing: `signatureNode` is the signature under construction, already resolved into this
+    // subtree. It has no SignatureValue yet, so remove exactly that element.
+    if (utils.isDescendantOf(signatureNode, node) && signatureNode.parentNode) {
+      signatureNode.parentNode.removeChild(signatureNode);
+      return node;
+    }
+    // Verifying: `signatureNode` was loaded from a separate parse, so find the matching
+    // signature in this subtree by its SignatureValue.
```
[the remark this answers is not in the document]
cjbarth: @shunkica, do you think there is merit to this finding?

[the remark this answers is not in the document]
shunkica: It was pre-existing but I added a throw here now.

PR #526, opened 2025-12-28
fix: Support nested enveloped signature location (#525)

### LINKED ISSUE #525: [BUG] EnvelopedSignature transform fails when signature is nested in child element
# Description

The library produces invalid signatures when using the enveloped-signature transform with nested signature placement using location.

# Reproduction

```
import { SignedXml } from "../src/index";
import { generateKeyPairSync } from "crypto";
import { expect } from "chai";
import * as xpath from "xpath";
import * as xmldom from "@xmldom/xmldom";
import * as isDomNode from "@xmldom/is-dom-node";

describe("EnvelopedSignature with nested signature location", function () {
  it("should verify signature when signature is nested in a child element", function () {
    // Generate keys
    const { privateKey, publicKey } = generateKeyPairSync("rsa", {
      modulusLength: 2048,
      publicKeyEncoding: { type: "spki", format: "pem" },
      privateKeyEncoding: { type: "pkcs8", format: "pem" },
    });

    const xml = "<root><SignatureInformation/></root>";

    // Sign
    const sig = new SignedXml();
    sig.privateKey = privateKey;

    sig.addReference({
      xpath: "/*",
      transforms: [
        "http://www.w3.org/2000/09/xmldsig#enveloped-signature",
        "http://www.w3.org/2001/10/xml-exc-c14n#",
      ],
      digestAlgorithm: "http://www.w3.org/2001/04/xmlenc#sha256",
    });

    sig.canonicalizationAlgorithm = "http://www.w3.org/2001/10/xml-exc-c14n#";
    sig.signatureAlgorithm = "http://www.w3.org/2001/04/xmldsig-more#rsa-sha256";

    // Place signature inside SignatureInformation (not as direct child of root)
    sig.computeSignature(xml, {
      location: {
        action: "append",
        reference: "//*[local-name()='SignatureInformation']",
      },
    });

    const signedXml = sig.getSignedXml();

    // Verify
    const doc = new xmldom.DOMParser().parseFromString(signedXml);
    const signatureNode = xpath.select1(
      "//*[local-name(.)='Signature' and namespace-uri(.)='http://www.w3.org/2000/09/xmldsig#']",
      doc,
    );
    isDomNode.assertIsNodeLike(signatureNode);

    const verifier = new SignedXml();
    verifier.publicCert = publicKey;
    verifier.loadSignature(signatureNode);

    const result = verifier.checkSignature(signedXml);

    expect(result, "Signature verification should succeed").to.be.true;
  });
});
```

shunkica:
## Description

This PR fixes an issue where the enveloped signature transformation was unable to find and remove Signature elements that were nested within other elements in the document.

## Changes

- ~~Updated the XPath query in  from  to  to search for Signature elements at any depth, not just direct children~~
- ~~Added a test case demonstrating the issue with nested signatures~~
- `computeSignature` sets `signatureNode` before processing references, and restores it if reference processing throws
- `getCanonXml` maps the signature element into the cloned subtree so the enveloped-signature transform removes exactly the signature being created, at any depth
- `EnvelopedSignature` removes the mapped node by identity during signing; verification still matches by `SignatureValue`
- Tests for nested locations, counter-signing at `/root` and `/root/container`, tampering, and state recovery after a failed sign

## Related Issue

Fixes #525

## Testing

Added a test case that verifies signatures can be validated when the Signature element is nested within other elements in the document structure.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved XML signature handling for signatures located within nested elements.
  * Fixed canonicalization so the intended signature is removed while nested signatures remain intact.
  * Improved recovery when signature generation or reference processing fails.
  * Ensured signatures remain verifiable across subsequent signing operations and correctly detect tampering.

* **Tests**
  * Added coverage for nested signature locations, canonicalization behavior, repeated signing, and tampered signatures.

### DIFF
--- a/src/enveloped-signature.ts
+++ b/src/enveloped-signature.ts
@@ -18,7 +18,7 @@ export class EnvelopedSignature implements CanonicalizationOrTransformationAlgor
   process(node: Node, options: CanonicalizationOrTransformationAlgorithmProcessOptions): Node {
     if (null == options.signatureNode) {
       const signature = xpath.select1(
-        ".//*[local-name(.)='Signature' and namespace-uri(.)='http://www.w3.org/2000/09/xmldsig#']",
+        "./*[local-name(.)='Signature' and namespace-uri(.)='http://www.w3.org/2000/09/xmldsig#']",
         node,
       );
       if (isDomNode.isNodeLike(signature) && signature.parentNode) {
--- a/src/enveloped-signature.ts
+++ b/src/enveloped-signature.ts
@@ -17,7 +17,7 @@ export class EnvelopedSignature implements CanonicalizationOrTransformationAlgor
   process(node: Node, options: CanonicalizationOrTransformationAlgorithmProcessOptions): Node {
     if (null == options.signatureNode) {
       const signature = xpath.select1(
-        "./*[local-name(.)='Signature' and namespace-uri(.)='http://www.w3.org/2000/09/xmldsig#']",
+        ".//*[local-name(.)='Signature' and namespace-uri(.)='http://www.w3.org/2000/09/xmldsig#']",
         node,
       );
       if (isDomNode.isNodeLike(signature) && signature.parentNode) {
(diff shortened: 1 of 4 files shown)


cjbarth:
@shunkica , I've fixed a few regressions here. I think this is ready to merge. What do you think?

shunkica:
LGTM. Updated PR body.

cjbarth (approved):

PR #528, opened 2026-01-16
Fix double callback invoke on unhandled exception

### LINKED ISSUE #527: Callback invoked twice on unhandled exception
The callback handler is invoked twice when there is an unhandled exception in the callback, eg:

```js
const { SignedXml } = require("xml-crypto");
const crypto = require("crypto");

const xml = `<x xmlns:wsu='http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-utility-1.0.xsd' Id='_1'></x>`;
const sig = new SignedXml();
sig.privateKey = crypto.generateKeyPairSync("rsa", {
  modulusLength: 2048,
  privateKeyEncoding: {
    type: "pkcs1",
    format: "pem"
  }
}).privateKey;
sig.canonicalizationAlgorithm = "http://www.w3.org/2001/10/xml-exc-c14n#";
sig.signatureAlgorithm = "http://www.w3.org/2000/09/xmldsig#rsa-sha1";
sig.computeSignature(xml, (err, result) => {
  console.log("Callback invoked", !!err, !!result);
  throw new Error('Error Thrown');
});

// Results in:
// Callback invoked false true
// Callback invoked true false

// Error: Error Thrown
```

The callback should only be invoked a single time, eg:

```js
require('fs').readdir(".", (err, files) => {
  console.log("Callback invoked", !!err, !!files);
  throw new Error('Error Thrown');
});

// Results in:
// Callback invoked false true

// Error: Error Thrown
```

I believe the problem is the synchronous execution of the callback in https://github.com/node-saml/xml-crypto/blob/master/src/types.ts#L259

adamjmcgrath:
Reported and diagnosed by @adamjmcgrath in #527: when a caller's callback throws, `computeSignature(xml, cb)` invokes it a second time, passing the callback's own error back as `err`.

```
Callback invoked false true
Callback invoked true false     <- should not happen
Error: Error Thrown
```

## Cause

`createOptionalCallbackFunction` invoked the callback *inside* the `try`, so an exception thrown by the callback landed in the `catch` and was handed straight back to it:

```ts
try {
  const result = syncVersion(...args);
  possibleCallback(null, result);   // throws here...
} catch (err) {
  possibleCallback(err ...);        // ...and is caught here, invoking the callback again
}
```

Present since the helper was introduced in #343, so every release from v4.0.0 onward.

## Fix

Narrow the `try` to cover only `syncVersion`. Once the callback is outside it, its exceptions cannot re-enter the `catch`:

```ts
let result: T;
try {
  result = syncVersion(...args);
} catch (err) {
  possibleCallback(err instanceof Error ? err : new Error("Unknown error"));
  return;
}
possibleCallback(null, result);
```

The `return` is enforced by the compiler rather than by discipline — without it `result` is not definitely assigned and `tsc --strict` rejects the file, so the error path cannot regress into a double call.

## Why not `process.nextTick`

The original version of this PR deferred the success callback with `process.nextTick`. That also stops the double invocation, but it changes *when* the callback runs, and `computeSignature(xml, cb)` was effectively synchronous. Measured on the same input, the one line differing:

| | callback invocations | `getSignedXml()` immediately after |
|---|---|---|
| before | 2 | 780 |
| `process.nextTick` | 1 | **0** |
| this fix | 1 | 780 |

An existing caller reading `getSignedXml()` after `computeSignature(xml, cb)` would get an empty string rather than an exception — a silent breaking change to a semver-bound public API, in a library where the failure surfaces downstream as an unsigned document. Narrowing the `try` fixes the reported bug with byte-identical timing instead.

De-Zalgoing these callbacks is still worth doing, but as a deliberate major rather than inside a bug fix. Tracked for 7.0 in #546, alongside #545.

## Tests

One regression test driving the public `computeSignature(xml, callback)` path from the issue. It was watched failing first against the unfixed helper, for the reported reason:

```
AssertionError: expected [ null, 'Error Thrown' ] to deeply equal [ null ]
```

It is fully synchronous, so it no longer removes and restores the process `uncaughtException` listeners — the earlier version leaked mocha's handler for the rest of the run whenever it failed.

`npm run build`, `npm test` (219 passing) and `npm run lint` all clean.

fixes #527

## Summary by CodeRabbit

- **Bug Fixes**
  - Error-first callbacks now receive successful results synchronously.
  - Callback errors no longer cause the callback to be invoked a second time.
  - Error callbacks return immediately after being called, providing more predictable completion behavior.

- **Tests**
  - Added coverage confirming synchronous XML signature computation invokes the callback exactly once with no error.
  - Added validation that callback-thrown errors propagate correctly without duplicate callback execution.


cjbarth:
@adamjmcgrath , I've made some changes. What do you think?

markstos (changes requested):
I agree this is a bug and agree with the fix. Test coverage also looks good.

The added comments are unnecessary, though. 

In most cases, comments that describe bugs that the code no longer has are not useful. Git history is useful to reviewing prior states of the code and history of code changes.

test/types-tests.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+      xpath: "//*[local-name(.)='x']",
+      digestAlgorithm: "http://www.w3.org/2000/09/xmldsig#sha1",
+      transforms: ["http://www.w3.org/2001/10/xml-exc-c14n#"],
+    });
+    sig.canonicalizationAlgorithm = "http://www.w3.org/2001/10/xml-exc-c14n#";
+    sig.signatureAlgorithm = "http://www.w3.org/2000/09/xmldsig#rsa-sha1";
+    return sig;
+  }
+
+  // https://github.com/node-saml/xml-crypto/issues/527
```
markstos: I'll go ahead and approve, though I don't think we need this comment either. 

Yes, sometimes it's useful for tests to reference issues for tricky edge cases, but test coverage to make sure exceptions work and callbacks aren't called twice is normal stuff.

markstos (dismissed):

test/types-tests.spec.ts
cjbarth: I'll get a PR up that updated the AGENTS.md file accordingly.

Issue #530, opened 2026-01-27
[ENHANCEMENT]: ECDSA Support


madaster97:
### Is your feature request related to a problem? Please describe...

Hi there, I am in need of a SAML testing toolkit that can create SAML Responses that are signed and encrypted with ECSDA (going off the definition from the [`xmldsig-core` spec](https://www.w3.org/TR/xmldsig-core/Overview_diff.html#sec-ECDSA)).

I noticed this library doesn't seem to have support for that! (only RSA algs). Is ECDSA support something you would be interested in adding?

### Describe teh solution you'd like...

I was able to monkey-patch in ECDSA support by just **leaning on the node.js crypto engine** to notice that my provided privateKey is an EC key instead of RSA. This kicked in by removing the `RSA-` in the call to `crypto.createSign` in [signed-xml.ts](https://github.com/node-saml/xml-crypto/blob/master/src/signed-xml.ts), so:
- Instead of `var signer = crypto.createSign("RSA-SHA256")`
- Calling `var signer = crypto.createSign("SHA256")`

I realise that seems a bit fragile, and there are ways that we could **pre-parse the provided key instead**. Do either of those approaches seem acceptable?

At the same time I could poke around at moving to [web-crypto](https://github.com/node-saml/xml-crypto/issues/477) instead.

PS - I have a bit of history with Applied Crypto. I cut my teeth on more [JWT](https://github.com/auth0/express-openid-connect/pull/291) work, but in my day job I have had to support / implement SAML a lot and am looking to update our systems from RSA to ECDSA.

### Describe the alternatives you've considered...

Right now this library is hard-coded NOT to support the `http://www.w3.org/2001/04/xmldsig-more#ecdsa-sha256` algorithm. So no alternatives unless there is a bring-your-own-alg option I missed?

madaster97:
FYI I am looking into this further and it looks like node.js outputs in ASN.1 encoded `r , s` instead of IEEE-P1363 `r || s` by default. Should be a quick fix using the `ieee-p1363` flag of `dsaEncoding` on [`crypto.sign`](https://nodejs.org/api/crypto.html#cryptosignalgorithm-data-key-callback)

Kudos to this [example dotnet implementation](https://www.scottbrady.io/c-sharp/ecdsa-xml-dotnet) with an example `SignatureValue`

[...]

PR #531, opened 2026-01-31
Add ECDSA Support (#530)

madaster97:
Adds the feature I requested in #530 , support for the following algorithms (both signing and verifying):

- ECDSA-SHA256 <http://www.w3.org/2001/04/xmldsig-more#ecdsa-sha256>
- ECDSA-SHA512 <http://www.w3.org/2001/04/xmldsig-more#ecdsa-sha512>

I worked backwards from this [example dotnet implementation](https://www.scottbrady.io/c-sharp/ecdsa-xml-dotnet) (had a sample ecdsa-sha256 document). I added that sample document in a unit test as external confirmation that we verify these signatures correctly, and then made paired *sign and verify - ecdsa* tests that show that our signatures verify correctly.

Thanks to @uladkasach, for this [stack overflow answer](https://stackoverflow.com/a/77162831/15073613) that showed how to call the crypto APIs correctly.

While I was here, I also touched up `example.js`, which was missing some config requirements for `computeSignature`, like:

- `sig.canonicalizationAlgorithm`
- `sig.signatureAlgorithm`
- `addReference > digestAlgorithm`
- `addReference > transforms`

## Summary by CodeRabbit

- **New Features**
  - Added support for ECDSA-SHA256 and ECDSA-SHA512 XML signatures.
  - Added examples for signing and validating XML with certificates.
  - Added an example command for running the local XML signing workflow.
  - Included a sample signed XML document.

- **Documentation**
  - Documented the newly supported ECDSA signature algorithms.

- **Tests**
  - Added coverage for ECDSA signing, verification, tampering, malformed signatures, and key compatibility.


src/signature-algorithms.ts
```diff
@@ -53,6 +53,36 @@ export class RsaSha256 implements SignatureAlgorithm {
...
+      // @ts-ignore
+      const res = signer.sign({ key: privateKey, dsaEncoding: 'ieee-p1363' }, "base64");
+
+      return res;
+    },
+  );
+
+  verifySignature = createOptionalCallbackFunction(
+    (material: string, key: crypto.KeyLike, signatureValue: string): boolean => {
+      const publicKey = crypto.createPublicKey(key);
```
ahacker1-securesaml, line 72: potential for key confusion attacks i.e. rsa public key being passed here

madaster97: I had looked into this, but the crypto functions seem to be validating this already. Do you mind demonstrating a key confusion attack?

ahacker1-securesaml: It's best practice to add checks (for this and the other signature-algorithms) that the verification key is the correct type (ec for this one). Currently there's no security vulnerability. If a rsa public key is passed in here, node:crypto will just verify with RSA (the key isn't converted to EC).

cjbarth:
@madaster97 , I've reviewed this and created some tests that are currently failing. Please have another look at this code and see what you can do to bring it up to standard. Also, please run `npm run lint:fix` before your next push to clean up the errors.

Issue #551, opened 2026-09-08
Audit the public export surface and deprecate internal helpers ahead of 7.0


cjbarth:
## Summary

The package publishes 40 symbols, and a good share of them are internal helpers that became public by accident rather than by design. Now that `src/index.ts` lists its exports explicitly instead of using `export * from`, we can see the whole surface and decide what belongs in it.

The goal for 7.0 is to shrink the surface. This issue covers the audit and the 6.x deprecation pass that has to come first — a name cannot simply vanish in 7.0 without consumers having seen a warning.

## How the surface got this wide

`src/index.ts` did:

```ts
export * from "./types";
export * from "./utils";
```

Every module-level export in `utils.ts` was therefore public automatically. Adding a helper there — even one intended purely for use by `signed-xml.ts` — silently widened the package's API, because a sibling module and an external consumer reach it through the same `export` keyword. That is how `findChilds` stayed published with no callers, and how each new helper joined the surface without anyone choosing to publish it.

## Current surface

40 symbols: 18 types and 1 value from `types.ts`, 16 values from `utils.ts`, plus `SignedXml` and the four canonicalization classes.

The 18 types are genuinely consumer-facing — dependents such as `node-saml` import `Reference`, `SignedXmlOptions` and friends directly. Expect that block to survive mostly intact. The `utils` block is where the work is:

| Export | Callers outside `utils.ts` | Notes |
| --- | --- | --- |
| `findChilds` | 0 | deprecated alias; see #550 |
| `isArrayHasLength` | 17 | internal predicate |
| `findChildren` | 16 | internal DOM helper |
| `findAttr` | 5 | internal DOM helper |
| `encodeSpecialCharactersInText` | 5 | internal serialization helper |
| `isDescendantOf` | 4 | internal DOM helper |
| `encodeSpecialCharactersInAttribute` | 3 | internal serialization helper |
| `validateDigestValue` | 3 | internal; constant-time comparison |
| `findAncestorNs` | 3 | plausibly useful to implementors |
| `EXTRACT_X509_CERTS` | 2 | internal regex |
| `derToPem` | 2 | plausibly useful to consumers |
| `pemToDer` | 2 | plausibly useful to consumers |
| `findAncestorNsForNode` | 2 | added recently; internal |
| `BASE64_REGEX` | 1 | internal regex |
| `PEM_FORMAT_REGEX` | 1 | internal regex |
| `normalizePem` | 1 | plausibly useful to consumers |

[...]

Issue #550, opened 2026-09-08
Remove deprecated findChilds() alias


cjbarth:
## Summary

`findChilds` is a deprecated alias of `findChildren` that survives only as a published export. Remove it in 7.0.

```ts
/** @deprecated */
export function findChilds(node: Node | Document, localName: string, namespace?: string) {
  return findChildren(node, localName, namespace);
}
```

It has no internal callers left. Its one remaining reference in `src/` is the re-export in `src/index.ts`; everything internal already calls `findChildren` (16 call sites). No test references it either.

## Why now

`src/index.ts` used to re-export it via `export * from "./utils"`, which never named the symbol, so the `deprecation/deprecation` ESLint rule stayed quiet. Now that the wildcard has been replaced with an explicit list, naming it trips the rule and the export needs a suppression to build:

```ts
findChildren,
// Deprecated alias of `findChildren`, still published because `export *` did.
// eslint-disable-next-line deprecation/deprecation
findChilds,
```

That suppression is the only reason the name still compiles into the surface, and it should go away with the function.

## Scope

- Delete `findChilds` from `src/utils.ts`.
- Delete the entry and its `eslint-disable` from the export list in `src/index.ts`.
- Note the removal in the changelog as a breaking change.

## Migration

`findChilds(node, localName, namespace)` → `findChildren(node, localName, namespace)`. The signature and behaviour are identical; the alias only ever forwarded.

## Before removing

Consider whether this needs a runtime deprecation warning in a 6.x release first, so consumers see it before the name disappears. The house pattern is in `src/signed-xml.ts`:

```ts
import { deprecate } from "util";

const warnX = deprecate(() => {}, "`x()` is deprecated and will be removed in a future version. Use `y()` instead.", "XML_CRYPTO_X");
```

Today `findChilds` carries a bare `/** @deprecated */` with no replacement named and no runtime warning, so a JavaScript consumer gets no signal at all. See the companion issue on auditing the rest of the public surface — this one is the clearest instance of the same problem.

Issue #497, opened 2025-04-15
v6.1.0 computeSignature() causes DeprecationWarning


strawberrycz:
**Issue Description**
v6.0.1 is OK, but in v6.1.0 I am getting DeprecationWarning when computeSignature() is called.

**Code Snippet**
```
private signXml(xml: string) {
    const sig = new SignedXml({
        privateKey: ERP_PRIV_KEY,
        getKeyInfoContent: () => {
            const certContent = ERP_CERT.toString()
                .match(/-----BEGIN CERTIFICATE-----([\s\S]*?)-----END CERTIFICATE-----/)?.
                ?.replace(/\n/g, '') || '';

            return `<X509Data><X509Certificate>${certContent}</X509Certificate></X509Data>`;
        },
        signatureAlgorithm: 'http://www.w3.org/2001/04/xmldsig-more#rsa-sha256',
        canonicalizationAlgorithm: 'http://www.w3.org/2001/10/xml-exc-c14n#',
    });

    sig.addReference({
        xpath: "//*[local-name()='ZalozeniPredpisuDotaz']",
        transforms: ['http://www.w3.org/2001/10/xml-exc-c14n#'],
        digestAlgorithm: 'http://www.w3.org/2001/04/xmlenc#sha256',
        uri: '',
        isEmptyUri: true,
    });

    sig.computeSignature(xml, {
        location: {
            reference: "//*[local-name()='ZalozeniPredpisuDotaz']",
            action: 'append',
        },
    });

    return sig.getSignedXml();
}
```

**Additional Context**
- Library version: xml-crypto@6.1.0
- Error messages: 
```
(node:720) DeprecationWarning: getReferences() is deprecated. Use `.getSignedReferences()` instead.
(Use `node --trace-deprecation ...` to show where the warning was created)

cjbarth:
We are aware and welcome PRs to address this.

cjbarth:
@ahacker1-securesaml , is it possible to provide a clear way to use NaCl Box for users. That link you sent doesn't have any instructions for use in a Node project. It also doesn't have directions on how to use it on Windows. If NaCl Box won't work for users of xml-crypto, and someone has to work with XML for project-specific reasons (SAML, SOAP, et.c), what alternative would you suggest?

cjbarth:
[...]

PR #559, opened 2026-09-09
chore: deprecate the internal helpers 7.0 withdraws

cjbarth:
Refs #550, #551 — the 6.x half, meant to land before `master` rolls over to 7.x.

#561 and #563 withdraw `findChilds` and the internal `utils` helpers. A name cannot vanish without consumers having seen a warning, so this marks the removal list `@deprecated` with a replacement named and, for everything that is a function, a `util.deprecate` runtime warning following the `getOriginalXmlWithIds()` pattern.

## Where the wrapping lives, and why

In `index.ts`, not `utils.ts`. Siblings reach these helpers as `utils.x`, which stays unwrapped, so no internal call path warns — warning on our own calls is what made #497 unpleasant. There is a test that runs a full verification and asserts none of these warnings fire.

## Covered

| Deprecated | Replacement named in the warning |
| --- | --- |
| `findChilds` | `findChildren()` |
| `findAttr`, `findChildren`, `isDescendantOf` | a DOM API or the `xpath` package |
| `isArrayHasLength` | `Array.isArray(x) && x.length > 0` |
| `encodeSpecialCharactersInAttribute`, `encodeSpecialCharactersInText` | an XML serializer |
| `validateDigestValue` | `crypto.timingSafeEqual(...)`, and never `===` |
| `BASE64_REGEX`, `EXTRACT_X509_CERTS`, `PEM_FORMAT_REGEX` | no replacement; internal parsing details |

`findChilds` previously carried a bare `/** @deprecated */` with no replacement named and no runtime signal, so a JavaScript consumer got nothing at all.

The three regexes can only carry the JSDoc tag — `util.deprecate` wraps a function and a `RegExp` has no call to intercept. The README says so rather than leaving it looking like an oversight.

## Not deprecated

`derToPem`, `pemToDer`, `normalizePem` and `findAncestorNs` stay in 7.0 and are left alone.

## Safe for a minor

Verified the runtime export surface is byte-for-byte identical to `master`'s, so nothing breaks. `@deprecated` will surface in editors and TypeScript builds, which is the point.

## Verification

`npm run build && npm test && npm run lint` clean; 251 passing (241 + 10).

## Summary by CodeRabbit

* **Documentation**
  * Added guidance on legacy utility exports, their replacements, and planned removal ahead of version 7.0.
  * Clarified which utility helpers remain supported.

* **Deprecations**
  * Legacy utility functions now emit deprecation warnings when called.
  * Deprecated regular-expression constants remain available with documentation notices.
  * Existing supported utility exports remain available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,27 @@
 
 ## Upgrading
 
+### Deprecated ahead of 7.0
+
+The package used to re-export everything in its internal `utils` module, so helpers written for
+`signed-xml.ts` became public API by accident. These are deprecated as of this release and will
+be removed in 7.0:
+
+| Deprecated                                                            | Instead                                                                                                                                             |
+| --------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `findAttr`, `findChildren`, `findChilds`, `isDescendantOf`            | use a DOM API, or [xpath](https://github.com/goto100/xpath)                                                                                         |
+| `encodeSpecialCharactersInAttribute`, `encodeSpecialCharactersInText` | these implement [c14n special-character normalization](https://www.w3.org/TR/xml-c14n#ProcessingModel); an XML serializer escapes for you           |
+| `isArrayHasLength`                                                    | `Array.isArray(x) && x.length > 0`                                                                                                                  |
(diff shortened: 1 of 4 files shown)


cjbarth:
Closed in favour of #567 — the same commits, opened from `cjbarth/xml-crypto` instead of a branch pushed directly to this repo by mistake. The branch here has been deleted.

PR #564, opened 2026-09-09
feat!: give asynchronous work its own entry points

### LINKED ISSUE #546: Design the sync/async model for 7.0 (Web Crypto, remote keys, and retiring the callback overloads)
## Summary

Decide how `xml-crypto` exposes synchronous and asynchronous operation in 7.0, and restructure the internals so both are supported without duplicating the signing and verification logic.

This is the design half of the Web Crypto work. #545 covers the type abstraction; this issue covers control flow. Neither should be settled in isolation, but they can be implemented separately.

## Why this needs deciding now

Three separate things want an asynchronous path, and today none of them has a working one:

- **Web Crypto** (#477, #513) — `crypto.subtle` is promise-only. There is no synchronous `subtle.sign()`, so any non-Node runtime forces async.
- **Remote keys** — HSM / KMS / signing-server implementations of `SignatureAlgorithm`. This is the documented reason the callback overloads exist.
- **`checkSignature`'s async path is currently broken.** `checkSignature` calls `verifySignature` in its three-argument synchronous form and never passes the callback down, so an async-only verifier returns `undefined` and a valid signature is reported as incorrect. Regressed in v6.1.0; still present. Fails closed, so it is a correctness bug rather than a hole, but the README documents a feature that does not work.

Meanwhile the synchronous path is the one nearly every consumer uses, and it must not get slower or become contagious.

### Promise overhead is not the constraint

```
RSA-2048 sign, sync           : 738.8 us
RSA-2048 sign, async-wrapped  : 753.3 us
bare await, no work           : 0.675 us
promise overhead as % of sign : 0.091%
```

An `await` is under a tenth of a percent of an RSA signature. The reason not to make everything async is **caller contagion**, not CPU: a consumer validating a SAML response inside a synchronous code path cannot call an async method without restructuring upward. That is an API cost and it is real, but it is not a performance cost.

### The dual-mode pattern is expensive internally

#513 keeps one function that is sync or async depending on whether a callback was passed. To make that work, every method on the call path had to be doubled:

```ts
private validateReference(ref, doc): boolean;
private validateReference(ref, doc, callback): void;
private createReferences(doc, prefix): string;
private createReferences(doc, prefix, callback): void;
private createSignedInfo(doc, prefix): string;
private createSignedInfo(doc, prefix, callback): void;
```

each with `// Sync mode` / `// Async mode` branches, plus a hand-rolled sequential accumulator for the reference loop. `signed-xml.ts` grows by +340/−114, and most of that is bridging rather than function. It also replicates the seam that produced #527, five more times.

### "Sync unless you provide a callback" is observable to the caller

One piece of caller code, two algorithms, dual-mode semantics:

```
[A] Node algorithm  (sync)
   callback ran. audit is: null          <- handler runs before the caller finishes setup
   caller finished setup.

[B] WebCrypto algo  (async)
   caller finished setup.
   callback ran. audit is: {"signedAt":"now"}
```

and, more seriously:

```
[A] Node algorithm  -> try/catch CAUGHT: caller's handler failed
[B] WebCrypto algo  -> try/catch did NOT catch it
   ...it surfaced later as uncaughtException: caller's handler failed
```

Switching `signatureAlgorithm` — a one-line configuration change — silently changes whether the caller's own `try`/`catch` works, and whether state assigned after the call is visible to the handler. For a callback that carries "is this signature valid", that is not academic.

### What Node actually does

Node's error-first callbacks are **always** asynchronous, including on the error path and including calls it could answer immediately:

```
  caller finished setup.              <- always first
  readFile error   cb -> after is: set    (/definitely/not/here)
  dns.lookup cb       -> after is: set    (localhost, from /etc/hosts)
  readFile success cb -> after is: set
```

Node's answer to sync-vs-async is a pair of separately named functions — `readFile` / `readFileSync`, `writeFile` / `writeFileSync`, and roughly forty more. "Sync unless a callback is provided" is the thing that convention exists to avoid.

### 1. Write the logic once: sync phases separated by async barriers

The asynchronous work is confined to three leaf operations — `getHash`, `getSignature`, `verifySignature` — across six call sites in `signed-xml.ts`. Everything around them (canonicalization, XPath, digest comparison, DOM assembly) is synchronous computation with nothing to await. So each flow is a sequence of synchronous phases with exactly two async barriers:

```
sign:   canonicalize refs -> [hash each] -> assemble + canonicalize SignedInfo -> [sign] -> insert & serialize
verify: canonicalize refs -> [hash each] -> compare digests, canonicalize SignedInfo -> [verify] -> set signedReferences
```

Write the phases once, synchronous, with no modes. Then two thin orchestrators:

```ts
computeSignature(xml: string): void {
  const refs   = this.canonicalizeReferences(doc);           // shared, sync
  const digest = refs.map((r) => hash.getHash(r.canon));     // sync leaf
  const info   = this.assembleSignedInfo(digest);            // shared, sync
  this.finalize(algo.getSignature(info, this.privateKey));   // sync leaf + shared
}

async computeSignatureAsync(xml: string): Promise<SignedXml> {
  const refs   = this.canonicalizeReferences(doc);                          // same fn
  const digest = await Promise.all(refs.map((r) => hash.getHashAsync(r.canon)));
  const info   = this.assembleSignedInfo(digest);                           // same fn
  this.finalize(await algo.getSignatureAsync(info, this.privateKey));       // same fn
  return this;
}
```

The mode lives at the top where it is visible, rather than threaded through five private methods.

### 2. Optional async twins on the algorithm interfaces

```ts
interface HashAlgorithm {
  getAlgorithmName(): HashAlgorithmType;
  getHash(xml: string): string;
  getHashAsync?(xml: string): Promise<string>;
}
```

Same shape for `SignatureAlgorithm`. Implementations pick a side: a Node-backed algorithm implements the synchronous method, a Web Crypto-backed one implements the asynchronous method. Nobody writes both, so this is not a doubled API for implementers.

### 3. Fail closed at the wrong entry point

Calling `computeSignature()` with an async-only algorithm configured throws immediately and says what to do:

> `WebCryptoSha256 is async-only; use computeSignatureAsync()`

#513 gets this instinct right for hashes. With separate methods the type system carries most of it and the guard covers the JavaScript caller.

### 4. Retire the callback overloads

`createOptionalCallbackFunction` and the callback forms go. A caller who wants a callback can `.then()` or `promisify`. Removing them is a loud break — a failed compile or a `TypeError` — rather than the silent one that changing callback timing would produce.

If we would rather keep a callback form, it must be **always deferred**, matching Node. Dual-mode timing should not be designed in deliberately.

## Open questions

- Do the async entry points get their own names (`computeSignatureAsync`) or a separate namespace / class?
- Should `checkSignature`'s currently-broken async path be repaired in 6.x first, or left to be superseded by this work in 7.0?
- Do we need a bridging helper for implementers who genuinely want one algorithm class usable in both modes?
- Does anything other than the three crypto leaves need to become async? Canonicalization and XPath are believed to be entirely synchronous — worth confirming before committing to the phase structure.
- Error semantics: `computeSignature(xml, cb)` today throws synchronously for `signatureAlgorithm is required` and `Private key is required`, but calls back for `location.action`. In the promise API every failure becomes a rejection. Confirm that is what we want.

## References

- #545 — `BinaryLike` / `KeyLike` type abstraction (the other half of this work)
- #477 — Web Crypto support request
- #513 — draft Web Crypto PR; source of the dual-mode cost measurement
- #527 — callback invoked twice; the bug this seam produced in 6.x

cjbarth:
Closes #546

Three things want an asynchronous path and none of them had a working one: Web Crypto (`crypto.subtle` is promise-only), remote keys in an HSM or KMS, and `checkSignature`'s own callback form.

That last one is worth stating plainly — it did not work at all. `checkSignature` called `verifySignature` in its three-argument synchronous shape and never passed the callback down. Reproduced on `master` with an async-only verifier:

```
Error: sync not supported
    at AsyncOnlyRsaSha1.verifySignature (bug546.js:21:11)
    at SignedXml.checkSignature (lib/signed-xml.js:250:31)
```

A verifier that answers only through its callback cannot report a valid signature. `checkSignatureAsync` replaces that path, and there is a test for it citing the issue.

## New entry points

`computeSignatureAsync`, `checkSignatureAsync`, and `validateElementAgainstReferencesAsync` join their synchronous counterparts. `computeSignatureAsync` resolves with the instance so `getSignedXml()` can be chained.

## Written once

Only three operations can be asynchronous, so each flow is a sequence of synchronous phases with two barriers, and the mode lives at the top of two thin orchestrators rather than being threaded through five private methods:

```
sign:   prepareSignature -> collectReferenceDigests -> [hash] ->
        canonicalize SignedInfo -> [sign] -> finalizeSignature

verify: prepareVerification -> locateReferences -> [hash] ->
        compare digests -> [verify] -> conclude
```

`collectReferenceDigests` builds the `Reference` elements and leaves each `DigestValue` empty. `validateReference` splits into `locateReference` (resolve and canonicalize) and `acceptReferenceDigest` (compare, record signed content). `validateElementAgainstReferences` becomes two thin loops over one lazy generator, so a caller that matches on the first reference still does no work for the rest.

Everything between the barriers is shared, so the two entry points produce identical output. Asserted for all four bundled signature algorithms; the pre-existing byte-exact expected-XML test now runs through `computeSignatureAsync` and still matches to the byte.

## Optional async twins

`getHash`, `getSignature` and `verifySignature` become optional and gain `getHashAsync`, `getSignatureAsync`, `verifySignatureAsync`. An implementation provides whichever forms its backend supports and no more — nobody writes both, which is what the issue asks for. Making the synchronous methods optional is what lets an async-only algorithm exist without a throwing stub; the guards below make the failure legible instead.

The asynchronous entry points fall back to a synchronous method, so they accept every algorithm the synchronous ones do, the bundled `node:crypto` algorithms included.

## Fail closed at the wrong entry point

```
WebCryptoSha256 is async-only; use computeSignatureAsync()
AsyncOnlyRsaSha256 is async-only; use checkSignatureAsync()
NeitherFormSha256 implements neither getHash() nor getHashAsync()
```

## Retiring the callbacks

`createOptionalCallbackFunction` and `ErrorFirstCallback` are gone. "Sync unless you pass a callback" was observable: switching `signatureAlgorithm` changed whether the caller's own `try`/`catch` caught a handler's error and whether state assigned after the call was visible to the handler. Node's answer to this is a pair of separately named functions.

The issue asks for the break to be loud. For TypeScript it is a compile error. For JavaScript it would have been silent — signing successfully and never calling back — so `computeSignature` and `checkSignature` reject a function argument:

```
TypeError: The callback form was removed in 7.0; use computeSignatureAsync(), which returns a promise
```

## Open questions from the issue

- **Naming** — `*Async` methods on `SignedXml`, not a separate class.
- **Fix `checkSignature`'s async path in 6.x first?** Superseded here rather than repaired; the callback form it belonged to is gone.
- **Bridging helper for implementers?** Not added. The asynchronous entry points already accept a synchronous algorithm, which covers the case.
- **Anything else needing to be async?** No. Canonicalization and XPath are synchronous throughout; the six crypto call sites were the only ones.
- **Error semantics** — every failure in the promise API is a rejection, including configuration errors that throw synchronously today. There is a test asserting `computeSignatureAsync` with no `signatureAlgorithm` rejects rather than throwing before the promise settles.

## Not changed

Reference digests are still checked before the `SignedInfo` signature. The `TODO` about reversing that order is a separate behaviour change and stays a `TODO`.

## Verification

`npm run build && npm test && npm run lint` clean; 257 passing (241 + 16). All pre-existing tests pass unmodified apart from the one that used the callback form.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,48 @@
 
 ## Upgrading
 
+### Upgrading to 7.0
+
+Asynchronous work is done through its own entry points rather than by passing a callback.
+`computeSignature(xml, callback)`, `computeSignature(xml, options, callback)` and
+`checkSignature(xml, callback)` are gone, along with the `createOptionalCallbackFunction`
+helper and the `ErrorFirstCallback` type:
+
+```js
+// before
+sig.computeSignature(xml, (err) => {
+  if (err) return done(err);
+  use(sig.getSignedXml());
+});
+
+// after
+const signedXml = (await sig.computeSignatureAsync(xml)).getSignedXml();
+```
+
+Passing a callback now throws a `TypeError` naming the replacement, rather than signing
+successfully and never calling back.
+
+"Synchronous unless you pass a callback" was observable to the caller: switching
+`signatureAlgorithm` changed whether the caller's own `try`/`catch` caught a handler's error and
+whether state assigned after the call was visible to the handler. Node's own answer to this is a
+pair of separately named functions, and that is what this is.
+
+`checkSignature`'s asynchronous path did not work at all: it called `verifySignature` in its
+three-argument synchronous form and never passed the callback down, so an async-only verifier
+could not report a valid signature. `checkSignatureAsync` replaces it.
+
+`HashAlgorithm.getHash`, `SignatureAlgorithm.getSignature` and
+`SignatureAlgorithm.verifySignature` are now optional, joined by `getHashAsync`,
+`getSignatureAsync` and `verifySignatureAsync`. An implementation provides whichever forms its
+backend supports — see [asynchronous signing and verification](#asynchronous-signing-and-verification).
+Existing synchronous implementations keep working unchanged; TypeScript code that _calls_ these
+methods through the interface type now has to account for them being optional.
+
+`validateElementAgainstReferences` gained an asynchronous twin,
+`validateElementAgainstReferencesAsync`.
+
+### Upgrading to 6.0
+
 The `.getReferences()` AND the `.references` APIs are deprecated.
 Please do not attempt to access them. The content in them should be treated as unsigned.
 
@@ -278,6 +320,7 @@ To sign xml documents:
     - `attrs` - a hash of attributes and values `attrName: value` to add to the signature root node
     - `location` - customize the location of the signature, pass an object with a `reference` key which should contain a XPath expression to a reference node, an `action` key which should contain one of the following values: `append`, `prepend`, `before`, `after`
     - `existingPrefixes` - A hash of prefixes and namespaces `prefix: namespace` that shouldn't be in the signature because they already exist in the xml
+- `computeSignatureAsync(xml, [options])` - as `computeSignature`, but returns a promise for this instance and awaits any [algorithm that can only work asynchronously](#algorithms-that-can-only-work-asynchronously)
 - `getSignedXml()` - returns the original xml document with the signature in it, **must be called only after `computeSignature`**
 - `getSignatureXml()` - returns just the signature part, **must be called only after `computeSignature`**
 - `getOriginalXmlWithIds()` - **[deprecated]** returns the original xml with Id attributes added on relevant elements, **must be called only after `computeSignature`**. Use the `location` option of `computeSignature()` to place the signature, then `getSignedXml()`. See [how to specify the location of the signature](#how-to-specify-the-location-of-the-signature).
@@ -287,6 +330,7 @@ To verify xml documents:
 - `loadSignature(signatureXml)` - loads the signature where:
   - `signatureXml` - a string or node object (like an [xmldom](https://github.com/xmldom/xmldom) node) containing the xml representation of the signature
 - `checkSignature(xml)` - validates the given xml document and returns `true` if the validation was successful
+- `checkSignatureAsync(xml)` - as `checkSignature`, but returns a promise and awaits any [algorithm that can only work asynchronously](#algorithms-that-can-only-work-asynchronously)
 
 ## Customizing Algorithms
 
@@ -332,11 +376,20 @@ function MySignatureAlgorithm() {
     return "signature of signedInfo as base64...";
   };
 
+  /*verify the given signature over the given material. return a boolean*/
+  this.verifySignature = function (material, key, signatureValue) {
+    return true;
+  };
+
   this.getAlgorithmName = function () {
     return "http://mySigningAlgorithm";
   };
 }
 ```
+
+If the backend cannot answer synchronously — Web Crypto, an HSM, a KMS, a signing server —
+implement `getHashAsync`, `getSignatureAsync` and `verifySignatureAsync` instead. See
+[asynchronous signing and verification](#asynchronous-signing-and-verification).
 
 Custom transformation algorithm.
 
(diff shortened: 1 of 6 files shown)


cjbarth:
Closed in favour of #571 — the same commits, opened from `cjbarth/xml-crypto` instead of a branch pushed directly to this repo by mistake. The branch here has been deleted.

Issue #538, opened 2026-08-10
`findNSPrefix` returns only the first xmlns declaration → duplicate default xmlns in non-exclusive C14N when subset declares any `xmlns:*`


msheby:
## Summary

`findNSPrefix` in `src/utils.ts` returns only the *first* xmlns-like attribute on the referenced subset root. `findAncestorNs` then filters ancestor namespaces by that single prefix. If the subset root declares any prefixed namespace (`xmlns:enc="..."`, `xmlns:xsi="..."`, etc.) **and** inherits a default namespace from an ancestor, the inherited default (prefix `""`) is not filtered from `ancestorNamespaces`. During non-exclusive C14N (`http://www.w3.org/TR/2001/REC-xml-c14n-20010315[#WithComments]`), the default namespace is then emitted twice — once via `renderNs`'s `defaultNs !== currNs` branch, once via the ancestorNamespaces merge — producing a duplicate `xmlns="..."` and a digest that no spec-compliant implementation will match.

## Impact

Any signature whose referenced subset declares a prefixed namespace and inherits a default namespace fails to interoperate. We hit this in a SMPTE standard where the referenced `<AuthenticatedPrivate>` element carries `xmlns:enc="http://www.w3.org/2001/04/xmlenc#"` — signatures produced by other conformant signers verify against each other but not against xml-crypto, and vice versa.

## Repro (Node 20, xml-crypto 6.1.2)

```js
import { DOMParser } from '@xmldom/xmldom'
import xmlCrypto from 'xml-crypto'
const { C14nCanonicalization, SignedXml } = xmlCrypto

const xml = `<Root xmlns="http://example.com/root">
  <Body Id="B" xmlns:enc="http://www.w3.org/2001/04/xmlenc#">
    <enc:CipherValue>x</enc:CipherValue>
  </Body>
</Root>`

const doc = new DOMParser().parseFromString(xml, 'text/xml')
const body = doc.getElementsByTagName('Body')[0]
const ancestors = SignedXml.findAncestorNs(doc, "//*[local-name()='Body']")

console.log('ancestorNamespaces:', ancestors)
// Actual:   [{ prefix: '', namespaceURI: 'http://example.com/root' }]
// Expected: []  (Body already inherits the default; nothing to hoist)

const c14n = new C14nCanonicalization().process(body, {
  ancestorNamespaces: ancestors
})
console.log(c14n)
[...]

PR #565, opened 2026-09-09
test: pin the hoisted ancestor namespace behaviour #541 fixed

cjbarth:
Refs #538, #541.

Three branches were open against defects in the ancestor-namespace path of non-exclusive C14N:

1. a hoisted ancestor default namespace not becoming the default its descendants were canonicalized against, so a child's own `xmlns=""` read as redundant and was dropped;
2. an ancestor declaration being appended alongside one the node made itself, producing two `xmlns` attributes on one element;
3. `process()` marking ancestor prefixes in scope before anything had been rendered, so the apex's own redeclaration of such a prefix was skipped as redundant and resolved to the outer binding.

**All three are already fixed on `master`.** #541 got there by a different route — `localDefaultNs`/`nodeDefaultNs` in `renderNs`, `newDefaultNs` seeded from the hoisted entry in the ancestor merge, and `findSubsetNSPrefixes` collecting every prefix the subset declares rather than the first. So the fixes those branches carried are obsolete and are dropped.

Their tests are not obsolete. `master` satisfies these spec requirements without asserting any of them, in an area that has now produced four separate defects.

## These are guards, not documentation

Each of the six unit cases fails against `src/c14n-canonicalization.ts` as it stood before #541 and passes on `master`. Two spot checks confirm they bite against plausible regressions:

| Perturbation of `master` | Failures |
| --- | --- |
| drop `newDefaultNs = ancestorNamespace.namespaceURI` from the ancestor merge | 7 (4 without this PR) |
| revert `findSubsetNSPrefixes` to the first declaration only | 12 (11 without this PR) |

The cases are parameterized over both `C14nCanonicalization` and `C14nCanonicalizationWithComments`, which the original branches did not do.

## The integration case

The one worth having. A hoisted default namespace leaking into a descendant leaves the signature verifying while `getSignedReferences()` — the API the README designates as the secure replacement for `getValidatedNode()` — reports the element in a namespace the sender never signed it in. A caller that dispatches on element namespace then acts on an identity that was never signed, and the sender picks it. Nothing else in the suite covers that.

Cited inline: C14N 1.0 [§2.3](https://www.w3.org/TR/xml-c14n/#ProcessingModel), the [§4.7](https://www.w3.org/TR/xml-c14n/#PropagateDefaultNSDecl) rationale for preserving `xmlns=""`, and [Namespaces in XML §2.2](https://www.w3.org/TR/REC-xml-names/#scoping) for prefix scoping.

## Milestone

v6.2, matching #541 — test-only, no behaviour change.

## Verification

`npm run build && npm test && npm run lint` clean; 254 passing (241 + 13; 6 cases × 2 canonicalizers + 1 integration).

### DIFF
--- a/test/c14n-non-exclusive-unit-tests.spec.ts
+++ b/test/c14n-non-exclusive-unit-tests.spec.ts
@@ -250,6 +250,70 @@ describe("C14N non-exclusive canonicalization tests", function () {
   });
 
   for (const Canonicalization of [C14nCanonicalization, C14nCanonicalizationWithComments]) {
+    describe(`${Canonicalization.name}: hoisted ancestor namespaces`, function () {
+      // A hoisted ancestor default namespace becomes the default the subset's descendants are
+      // canonicalized against. The rationale is spelled out in §4.7: the empty default is kept on
+      // output "so that e3 does not take on the default namespace qualification of e1".
+      // https://www.w3.org/TR/xml-c14n/#ProcessingModel
+      // https://www.w3.org/TR/xml-c14n/#PropagateDefaultNSDecl
+      it('renders xmlns="" on a descendant when the apex hoists an ancestor default namespace', function () {
+        test_C14nCanonicalization(
+          '<root xmlns="urn:A"><p:x xmlns:p="urn:p"><y xmlns=""></y></p:x></root>',
+          "//*[local-name()='x']",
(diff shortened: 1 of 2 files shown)


cjbarth:
Closed in favour of #572 — the same commits, opened from `cjbarth/xml-crypto` instead of a branch pushed directly to this repo by mistake. The branch here has been deleted.

PR #570, opened 2026-09-09
feat!: decouple public key and data types from node:crypto

### LINKED ISSUE #545: Decouple public key and data types from node:crypto (BinaryLike / KeyLike)
## Summary

Introduce runtime-neutral `BinaryLike` and `KeyLike` types in `src/types.ts`, and use them across the public API in place of `crypto.BinaryLike` / `crypto.KeyLike`.

```ts
export type BinaryLike = crypto.BinaryLike | ArrayBuffer;
export type KeyLike = crypto.KeyLike | CryptoKey | Uint8Array;
```

## Why

Today every public signature that touches key material or data to be signed is typed against `node:crypto` directly — `SignedXmlOptions.privateKey`, `SignedXmlOptions.publicCert`, `GetKeyInfoContentArgs.publicCert`, and both methods on the `SignatureAlgorithm` interface.

That hard-codes Node into the type surface. A caller cannot pass a `CryptoKey`, which is the only key representation the Web Crypto API produces, so the public types themselves block any Web Crypto backend before a line of implementation is written. This abstraction is a prerequisite for #477, and it is the piece of #513 that stands on its own regardless of what shape the rest of that work eventually takes.

## Why 7.0

Altering types exported from `src/types.ts` is breaking for consumers, so this belongs in a major even though it is additive in practice. Widening a parameter type is source-compatible for callers, but it is not compatible for anyone who *implements* `SignatureAlgorithm` — third-party HSM- and KMS-backed implementations exist, and their method signatures have to widen to match.

## Design note

In #513 these were introduced as blanket unions, which forced casts back out at every use site:

```ts
signer.update(signedInfo as crypto.BinaryLike);
const res = signer.sign(privateKey as crypto.KeyLike, "base64");
```

The cast is load-bearing: without it TypeScript correctly rejects the call (`TS2345: Type 'CryptoKey' is not assignable to parameter of type 'KeyLike | SignKeyObjectInput | SignPrivateKeyInput'`). So the widened type promises support that `RsaSha1`/`RsaSha256`/`RsaSha512`/`HmacSha1` do not have, and the cast deletes the diagnostic that would have said so.

What happens at runtime differs by arm, and the `CryptoKey` case is the concerning one:

- **`Uint8Array`** fails with `ERR_OSSL_UNSUPPORTED: error:1E08010C:DECODER routines::unsupported` — an opaque OpenSSL error a long way from the cause.
- **`CryptoKey`** does not fail at all. Node accepts it and emits `DEP0203 DeprecationWarning: Passing a CryptoKey to node:crypto functions is deprecated`. It appears to work, tests pass, and it breaks whenever Node removes the shim — and it can never work off-Node, where `node:crypto` does not exist. A Node-backed algorithm would look like it supports Web Crypto keys while doing nothing of the kind.

Options that avoid the cast rather than suppressing it:

1. Make the interface generic in its key type, so each implementation declares what it accepts.
2. Keep the union, but narrow through a helper that validates and throws a clear error instead of an `as`.
3. Separate interfaces per backend, so a Node algorithm never claims to take a `CryptoKey` at all.

Whichever way it goes, the goal is that each algorithm states what it actually accepts, rather than casting out of a union that promises more than it can take.

## Scope

In scope:
- Define `BinaryLike` / `KeyLike` in `src/types.ts`
- Apply them to `SignedXmlOptions`, `GetKeyInfoContentArgs`, and the `SignatureAlgorithm` interface
- Decide how the bundled Node-backed algorithms narrow to concrete types without blanket casts
- Document the change for implementers of `SignatureAlgorithm`

Out of scope — these get their own issues:
- Any Web Crypto algorithm implementation
- Async / callback API changes
- `HashAlgorithm` changes

## References

- #477 — Web Crypto support request
- #513 — draft Web Crypto PR these types were taken from

cjbarth:
Closes #545

Every public signature that touched key material named `node:crypto` directly, which hard-coded Node into the type surface: a caller could not pass a `CryptoKey` — the only key representation Web Crypto produces — so the types blocked a Web Crypto backend before any implementation existed.

`BinaryLike` and `KeyLike` are now defined in `src/types.ts` and used for `SignedXmlOptions.privateKey`/`publicCert`, `GetKeyInfoContentArgs.publicCert` and the `SignatureAlgorithm` interface.

## How the bundled algorithms narrow, without blanket casts

The issue offers three options; this takes the first. `SignatureAlgorithm` takes the accepted key type as a parameter, so each implementation states what it can really use:

| Algorithm | Declares |
| --- | --- |
| `RsaSha1`, `RsaSha256`, `RsaSha512`, `HmacSha1` | `SignatureAlgorithm<crypto.KeyLike \| Uint8Array>` |
| `RsaSha256Mgf1` | `SignatureAlgorithm<string \| Buffer>` — it needs a key it can put in a `SignPrivateKeyInput` |

Nothing casts back out. Verified that the compiler now rejects what it should:

```
TS2345: Argument of type 'CryptoKey' is not assignable to parameter of type
'Uint8Array<ArrayBufferLike> | KeyLike'.
```

The two arms the issue flagged as broken are now genuinely supported rather than narrowed away:

- **`Uint8Array`** is viewed as a `Buffer` (`Buffer.from(key.buffer, key.byteOffset, key.byteLength)`, no copy) instead of reaching OpenSSL as `ERR_OSSL_UNSUPPORTED`.
- **`ArrayBuffer`** data is viewed the same way, so `BinaryLike` is true for these algorithms. Test asserts an `ArrayBuffer` signs to the same bytes as the equivalent string.

## The residual hole, and why it fails closed

The signature algorithm is looked up by a URI read from the document under inspection, so a JavaScript caller can still pair a `CryptoKey` with a Node algorithm without the compiler seeing it. Confirmed on `master` that Node does not fail in that case — it signs:

```
master: signed with a CryptoKey, no error
(node:425927) [DEP0203] DeprecationWarning: Passing a CryptoKey to node:crypto functions is deprecated.
```

The signature verifies and the tests pass, so the algorithm appears to support a key it never really did, and it breaks whenever Node removes the shim. On this branch:

```
RsaSha256 needs a key that node:crypto accepts: a string, a Buffer, a Uint8Array, or a KeyObject
```

That check is the one piece of defensive code here; there is a comment saying why it exists despite the parameter type.

## Notes

- `CryptoKey` is spelled `crypto.webcrypto.CryptoKey` so it resolves under `lib: ["es2020"]` on the `@types/node@16` floor. A DOM or global `CryptoKey` is structurally identical and assignable.
- `getKeyInfoContent` still silently emits an empty `<X509Data/>` for key material it cannot read as PEM. That predates this change (a `KeyObject` does it today) and is left alone.

## Breaking

Source-compatible for callers. A break for anyone who **implements** `SignatureAlgorithm`: declare the key type, e.g. `implements SignatureAlgorithm<crypto.KeyLike>`. README `Upgrading` and a new "Declaring the key material your algorithm accepts" section cover it.

## Verification

`npm run build && npm test && npm run lint` clean; 244 passing (241 + 3). New tests in `test/key-material-tests.spec.ts` were confirmed to fail on `master` for the right reasons — two as compile errors, the `CryptoKey` one because `master` signs successfully.

## Summary by CodeRabbit

* **New Features**
  * Added broader type support for signing keys, verification certificates, and signed data, including `Uint8Array` and `ArrayBuffer`.
  * Signature algorithms can now declare the key formats they support.

* **Bug Fixes**
  * Added clear errors for unsupported key material instead of relying on inconsistent runtime behavior.
  * Normalized binary inputs to produce consistent signatures.

* **Documentation**
  * Updated key format, signing, verification, and custom algorithm guidance for version 7.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,23 @@
 
 ## Upgrading
 
+### Upgrading to 7.0
+
+Key material and data types no longer name `node:crypto` in the public API. `privateKey`,
+`publicCert` and the `SignatureAlgorithm` key arguments are now `KeyLike`, and the data to
+be signed is `BinaryLike`; both are exported from this package.
+
+This is source-compatible for callers. It is a break for anyone who **implements**
+`SignatureAlgorithm`: the interface now takes the accepted key type as a parameter, so an
+implementation should declare it — `implements SignatureAlgorithm<crypto.KeyLike>` for a
+`node:crypto`-backed one. See
+[declaring the key material your algorithm accepts](#declaring-the-key-material-your-algorithm-accepts).
+
+The bundled algorithms now throw when handed key material `node:crypto` cannot use, instead
+of letting a `CryptoKey` through Node's DEP0203 shim as if it were supported.
+
+### Upgrading to 6.0
+
 The `.getReferences()` AND the `.references` APIs are deprecated.
 Please do not attempt to access them. The content in them should be treated as unsigned.
 
@@ -61,8 +78,8 @@ signature algorithms enabled at same time.
 
 When signing a xml document you can pass the following options to the `SignedXml` constructor to customize the signature process:
 
-- `privateKey` - **[required]** a `Buffer` or pem encoded `String` containing your private key
-- `publicCert` - **[optional]** a `Buffer` or pem encoded `String` containing your public key
+- `privateKey` - **[required]** your private key, as a pem encoded `String`, a `Buffer`, a `Uint8Array`, or a `KeyObject`. Typed [`KeyLike`](#declaring-the-key-material-your-algorithm-accepts), which is wider than any one algorithm accepts
+- `publicCert` - **[optional]** your public certificate, in the same forms
 - `signatureAlgorithm` - **[required]** one of the supported [signature algorithms](#signature-algorithms). Ex: `sign.signatureAlgorithm = "http://www.w3.org/2001/04/xmldsig-more#rsa-sha256"`
 - `canonicalizationAlgorithm` - **[required]** one of the supported [canonicalization algorithms](#canonicalization-and-transformation-algorithms). Ex: `sign.canonicalizationAlgorithm = "http://www.w3.org/2001/10/xml-exc-c14n#WithComments"`
 
@@ -138,6 +155,8 @@ When verifying a xml document you can pass the following options to the `SignedX
 
 - `publicCert` - **[optional]** your certificate as a string, a string of multiple certs in PEM format, or a Buffer
 - `privateKey` - **[optional]** your private key as a string or a Buffer - used for verifying symmetrical signatures (HMAC)
+
+Both are typed [`KeyLike`](#declaring-the-key-material-your-algorithm-accepts).
 
 The certificate that will be used to check the signature will first be determined by calling `this.getCertFromKeyInfo()`, which function you can customize as you see fit. If that returns `null`, then `publicCert` is used. If that is `null`, then `privateKey` is used (for symmetrical signing applications).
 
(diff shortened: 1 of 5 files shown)


markstos:
@coderabbit review.

markstos (changes requested):
I agree with this direction.

I used codex to assist in reviewing this. I found two issues:

  - Runtime checks reject valid cross-realm inputs. toNodeData() (src/signature-
    algorithms.ts:13) uses instanceof ArrayBuffer, and toNodeKey() (src/signature-algorithms.ts:31)   uses instanceof Uint8Array. Both checks fail for values created in another realm, such as   node:vm, even though they satisfy the new public types. I reproduced:
      - Cross-realm ArrayBuffer reaches node:crypto unchanged and throws ERR_INVALID_ARG_TYPE.
      - Cross-realm Uint8Array key is rejected by toNodeKey().

    Use realm-independent brand detection.

  - [P2] The new migration example does not implement the documented interface. The class in   README.md:378 claims implements SignatureAlgorithm<crypto.KeyLike>, but getSignature returns nothing, does not provide the callback overload, and the class entirely omits verifySignature.  Consumers following the v7 migration documentation get compiler errors.

PR #576, opened 2026-09-11
fix: collapse Algorithm URIs per xsd:anyURI

### LINKED ISSUE #548: Algorithm identifiers are not whitespace-collapsed per xsd:anyURI, so line-wrapped Algorithm attributes fail verification
## Summary

Algorithm identifiers read out of a signature are matched against our registries (`CanonicalizationAlgorithms`, `SignatureAlgorithms`, `HashAlgorithms`) by **exact string equality**. We never apply the `xsd:anyURI` whitespace-collapse that the XML-DSIG schema calls for, so a signature whose `Algorithm` attribute was line-wrapped when it was signed — as it will be if the document is pretty-printed before signing — fails to verify with a misleading "is not supported" error naming an algorithm we very much do support.

XML attribute-value normalization (XML 1.0 §3.3.3) replaces a literal newline inside an attribute value with a space but does **not** trim or collapse it. So `Algorithm="http://...c14n-20010315\n          "` reaches us as the URI followed by eleven spaces, and the lookup misses.

This surfaces as an error message where the offending whitespace is essentially invisible, which makes it very hard for a user to diagnose. It is almost certainly what some of the reporters in #143 were actually hitting.

## Why this is correctness, not leniency

The XML Signature schema declares `Algorithm` as `xsd:anyURI`, and `xsd:anyURI` has a fixed `whiteSpace="collapse"` facet ([XML Schema Part 2 §4.3.6](https://www.w3.org/TR/xmlschema-2/#rf-whiteSpace)). A line-wrapped `Algorithm` is therefore well-formed and schema-valid, and its value is the collapsed URI. When we report `'http://...c14n-20010315           ' is not supported`, we are misreading a valid value, not rejecting a malformed one. The fix applies the schema's rule exactly: only XML whitespace, so NBSP is not stripped, and internal runs collapse to one space rather than disappearing. Nothing becomes an alias for a registered URI unless the schema already says it is that URI.

It is not a MUST we were violating, though. [XML-DSig 1.1 §7.1](https://www.w3.org/TR/xmldsig-core1/#sec-XML-1) notes that `Signature` is only laxly schema-valid, so a verifier may never see schema normalization. It therefore tells signers that a signature is only verifiable by other implementations if "attribute value white space be normalized". Signers that emit wrapped `Algorithm` attributes are outside that constraint, and rejecting them was permitted. Collapsing is the more faithful reading of the normative schema, and it removes an error that names an algorithm we support.

## Reproduction

Save as `repro.js` in the repo root and run `npm run build && node repro.js`. It signs a document with non-exclusive c14n, then re-verifies it four more times, each time line-wrapping exactly one `Algorithm` attribute and changing nothing else.

```js
const fs = require("fs");
const xpath = require("xpath");
const { DOMParser } = require("@xmldom/xmldom");
const { SignedXml } = require("./lib/index.js");

const C14N = "http://www.w3.org/TR/2001/REC-xml-c14n-20010315";
const ENVELOPED = "http://www.w3.org/2000/09/xmldsig#enveloped-signature";

// 1. Produce a valid signature using non-exclusive c14n.
const signer = new SignedXml({
  privateKey: fs.readFileSync("./test/static/client.pem"),
  canonicalizationAlgorithm: C14N,
  signatureAlgorithm: "http://www.w3.org/2000/09/xmldsig#rsa-sha1",
});
signer.addReference({
  xpath: "//*[local-name(.)='book']",
  transforms: [ENVELOPED, C14N],
  digestAlgorithm: "http://www.w3.org/2000/09/xmldsig#sha1",
});
signer.computeSignature("<library><book Id='b1'><title>Harry Potter</title></book></library>");
const signedXml = signer.getSignedXml();

// 2. Verify it, optionally after line-wrapping one Algorithm attribute the way a
//    pretty-printer would. Nothing else about the document changes.
function verify(label, xmlString) {
  const doc = new DOMParser().parseFromString(xmlString, "text/xml");
  const signatureNode = xpath.select1("//*[local-name(.)='Signature']", doc);
  const sig = new SignedXml({ publicCert: fs.readFileSync("./test/static/client_public.pem") });
  try {
    sig.loadSignature(signatureNode);
    console.log(`${label} -> checkSignature: ${sig.checkSignature(xmlString)}`);
  } catch (err) {
    console.log(`${label} -> THREW: ${err.message}`);
  }
}

const wrap = (needle) => signedXml.replace(needle, needle.replace(/"\/>$/, '\n          "/>'));

verify("as signed                ", signedXml);
verify("CanonicalizationMethod ws", wrap(`<CanonicalizationMethod Algorithm="${C14N}"/>`));
verify("Transform ws             ", wrap(`<Transform Algorithm="${C14N}"/>`));
verify("SignatureMethod ws       ", wrap('<SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/>'));
verify("DigestMethod ws          ", wrap('<DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/>'));
```

### Actual output (v6.0.0, `master` @ 8353bab)

```
as signed                 -> checkSignature: true
CanonicalizationMethod ws -> THREW: canonicalization algorithm 'http://www.w3.org/TR/2001/REC-xml-c14n-20010315           ' is not supported
Transform ws              -> THREW: canonicalization algorithm 'http://www.w3.org/TR/2001/REC-xml-c14n-20010315           ' is not supported
SignatureMethod ws        -> THREW: signature algorithm 'http://www.w3.org/2000/09/xmldsig#rsa-sha1           ' is not supported
DigestMethod ws           -> THREW: hash algorithm 'http://www.w3.org/2000/09/xmldsig#sha1           ' is not supported
```

### Expected output

Only the unmodified signature verifies. The repro wraps each `Algorithm` *after* signing, and those attributes are inside `SignedInfo`, so the other four must still fail. They should fail as a signature mismatch, not with a misleading `is not supported`:

```
as signed                 -> checkSignature: true
CanonicalizationMethod ws -> THREW: invalid signature: the signature value … is incorrect
Transform ws              -> THREW: invalid signature: the signature value … is incorrect
SignatureMethod ws        -> THREW: invalid signature: the signature value … is incorrect
DigestMethod ws           -> THREW: invalid signature: the signature value … is incorrect
```

A line-wrapped `Algorithm` should verify when the wrapping was present at signing time (pretty-print, then sign).

## Affected read sites

All four are in `src/signed-xml.ts`, and all take the raw `.value` with no normalization:

| Element | Location |
| --- | --- |
| `CanonicalizationMethod/@Algorithm` | `signed-xml.ts:633` (`loadSignature`) |
| `SignatureMethod/@Algorithm` | `signed-xml.ts:642` (`loadSignature`) |
| `DigestMethod/@Algorithm` | `signed-xml.ts:721` (`loadReference`) |
| `Transform/@Algorithm` | `signed-xml.ts:748` (`loadReference`) |

## Suggested fix

Normalize at **read time**, not inside the three `find*Algorithm` helpers.

The helpers are the tempting place to put a `.trim()`, but that would leave `this.canonicalizationAlgorithm` holding the polluted string, and it is compared by equality elsewhere. In particular, the exclusive-c14n guard in `loadSignature` (`signed-xml.ts:658-666`) does an `===` against the two `REC-xml-c14n-20010315` URIs to decide whether it can canonicalize `SignedInfo` in isolation. With a whitespace-laden value the guard doesn't recognize inclusive c14n, so the raw value goes on to `getCanonXml`. That is why the `CanonicalizationMethod ws` case above throws from inside `loadSignature` rather than from anywhere the user would think to look. It fails closed: it throws, and accepts nothing it shouldn't. Normalizing on the way in fixes the guard and the lookup together.

Collapse per `xsd:anyURI` — trim, and collapse internal whitespace runs to a single space — rather than a bare `.trim()`. A small shared helper in `src/utils.ts` applied at all four sites keeps this consistent.

## Security considerations

This is not a security fix: today these signatures are rejected, which fails closed.

Collapsing would weaken what we reject. Documents we refuse today would verify, and anything downstream that relies on that refusal would lose it. For example, a consumer that reads `Algorithm` from the DOM itself and bans an algorithm by exact string (say, `rsa-sha1`) would miss a wrapped value that we would then accept. That is a new disagreement between our reading of the document and the consumer's, and the consumer had no say in it.

[XML-DSig 1.1 §7.1](https://www.w3.org/TR/xmldsig-core1/#sec-XML-1) already puts the burden of normalizing attribute whitespace on signers, and no signer that emits wrapped `Algorithm` attributes has been reported. The case for widening what we accept is therefore weak.

## Suggested test coverage

- Each affected element (four total) covered by a signature that was **signed over** a line-wrapped `Algorithm` attribute, asserting it verifies. Ideally this comes from a fixture generated outside xml-crypto, so a canonicalization bug shared by our signer and verifier can't hide.
- A line-wrapped `Algorithm` added after signing still fails verification.
- Leading whitespace and internal whitespace runs, not just trailing.
- A negative test confirming a genuinely unregistered URI still throws `is not supported`, so the fix does not degenerate into fuzzy matching.

## Notes

Split out of #143. The original complaint there — that `http://www.w3.org/TR/2001/REC-xml-c14n-20010315` is unregistered — was fixed by #116 and shipped in v0.9.0 (2017-02-26), and non-exclusive c14n round-trips correctly on `master` today. This whitespace handling is the one live defect that still produces that exact error message for a supported URI.

wesleyschlenker:
**Maintainer review — 2026-09-11: hold this PR for a security regression.**

A reproduced case allows an attacker to insert an unsigned, whitespace-padded algorithm decoy into an already signed document and bypass a consumer's raw-string RSA-SHA1 prohibition. Master rejects the identifier; this PR normalizes it and returns `true` from `checkSignature`. This requires an existing valid RSA-SHA1 signature and the affected consumer check. The broad algorithm lookup predates this PR.

The [security review comment](https://github.com/node-saml/xml-crypto/pull/576#issuecomment-5639838236) records the tested revisions, results, prerequisites, controls, and why merely narrowing the library's selectors does not fully address the demonstrated consumer mismatch. For #548, preserve current rejection behavior and improve the diagnostic. Address signature structure and algorithm selection separately before reconsidering normalization. Earlier summaries that describe this change as having no merge-blocking security risk are superseded by that review.

---

Fixes #548

## Problem

`Algorithm` is looked up by exact string. XML 1.0 attribute-value normalization turns a line wrap into trailing spaces but does not collapse them, so a supported URI fails with `is not supported`. The exclusive-c14n guard in `loadSignature` compared that uncollapsed string too; the resulting unsupported-algorithm lookup failed closed.

## Fix

Collapse per `xsd:anyURI` (`whiteSpace="collapse"`) at the four `loadSignature` / `loadReference` read sites. Stored names then match the registries and the guard. The helper stays internal; the public barrel does not grow.

Changing an `Algorithm` attribute inside `SignedInfo` after signing still fails verification. The verify tests resign over the wrapped form, which is the pretty-print-then-sign path. However, the current descendant lookups can also select methods from unsigned content outside `SignedInfo`; normalization of those selected values enables the policy bypass described in the maintainer review above.

## Tests

- One verify case per affected element (`CanonicalizationMethod`, `Transform`, `SignatureMethod`, `DigestMethod`)
- Leading whitespace, internal whitespace runs, NBSP
- Unregistered URI still throws `is not supported`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved XML signature processing when algorithm identifiers contain leading, trailing, or repeated whitespace.
  * Signatures and references now resolve supported algorithm identifiers consistently when formatted with whitespace.
  * Unrecognized algorithm identifiers continue to be rejected after normalization.
  * Whitespace added after a signature was created still correctly causes verification to fail.

* **Tests**
  * Added coverage for whitespace handling across canonicalization, transformation, signature, and digest algorithms.

### DIFF
--- a/src/signed-xml.ts
+++ b/src/signed-xml.ts
@@ -630,7 +630,9 @@ export class SignedXml {
     }
 
     if (isDomNode.isAttributeNode(node)) {
-      this.canonicalizationAlgorithm = node.value as CanonicalizationAlgorithmType;
+      this.canonicalizationAlgorithm = utils.collapseAnyUri(
+        node.value,
+      ) as CanonicalizationAlgorithmType;
     }
 
     const signatureAlgorithm = xpath.select1(
@@ -639,7 +641,9 @@ export class SignedXml {
     );
 
     if (isDomNode.isAttributeNode(signatureAlgorithm)) {
-      this.signatureAlgorithm = signatureAlgorithm.value as SignatureAlgorithmType;
+      this.signatureAlgorithm = utils.collapseAnyUri(
+        signatureAlgorithm.value,
+      ) as SignatureAlgorithmType;
     }
 
     const signedInfoNodes = utils.findChildren(this.signatureNode, "SignedInfo");
@@ -718,7 +722,7 @@ export class SignedXml {
     if (!attr) {
       throw new Error(`could not find Algorithm attribute in node ${digestAlgoNode.toString()}`);
     }
-    const digestAlgo = attr.value;
+    const digestAlgo = utils.collapseAnyUri(attr.value);
 
     nodes = utils.findChildren(refNode, "DigestValue");
     if (nodes.length === 0) {
@@ -745,7 +749,7 @@ export class SignedXml {
         const transformAttr = utils.findAttr(transform, "Algorithm");
 
         if (transformAttr) {
-          transforms.push(transformAttr.value);
+          transforms.push(utils.collapseAnyUri(transformAttr.value));
         }
       }
 
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -31,6 +31,12 @@ export function findAttr(element: Element, localName: string, namespace?: string
   return null;
 }
 
+// `Algorithm` is xsd:anyURI. https://www.w3.org/TR/xmlschema-2/#anyURI
+// XML 1.0 §3.3.3 does not collapse, so this helper exists. https://www.w3.org/TR/xml/#AVNormalize
+export function collapseAnyUri(value: string): string {
+  return value.replace(/[\t\n\r ]+/g, " ").replace(/^ +| +$/g, "");
+}
+
 export function findChildren(node: Node | Document, localName: string, namespace?: string) {
   const element = (node as Document).documentElement ?? node;
   const res: Element[] = [];
--- a/test/algorithm-uri-whitespace.spec.ts
+++ b/test/algorithm-uri-whitespace.spec.ts
@@ -1,0 +1,144 @@
+import * as xpath from "xpath";
+import * as xmldom from "@xmldom/xmldom";
+import { SignedXml, findAncestorNs } from "../src/index";
+import * as fs from "fs";
+import { expect } from "chai";
+import * as isDomNode from "@xmldom/is-dom-node";
+
+const C14N = "http://www.w3.org/TR/2001/REC-xml-c14n-20010315";
+const ENVELOPED = "http://www.w3.org/2000/09/xmldsig#enveloped-signature";
+const RSA_SHA1 = "http://www.w3.org/2000/09/xmldsig#rsa-sha1";
+const SHA1 = "http://www.w3.org/2000/09/xmldsig#sha1";
+
+function signBook(): string {
+  const signer = new SignedXml({
+    privateKey: fs.readFileSync("./test/static/client.pem"),
+    canonicalizationAlgorithm: C14N,
+    signatureAlgorithm: RSA_SHA1,
+  });
+  signer.addReference({
+    xpath: "//*[local-name(.)='book']",
+    transforms: [ENVELOPED, C14N],
+    digestAlgorithm: SHA1,
+  });
+  signer.computeSignature("<library><book Id='b1'><title>Harry Potter</title></book></library>");
+  return signer.getSignedXml();
+}
+
+function wrapAlgorithm(xml: string, element: string, uri: string, wrapped: string): string {
+  const needle = `<${element} Algorithm="${uri}"/>`;
+  expect(xml.includes(needle), `signed xml should contain ${needle}`).to.equal(true);
+  return xml.replace(needle, `<${element} Algorithm="${wrapped}"/>`);
+}
+
+function load(xml: string): SignedXml {
+  const doc = new xmldom.DOMParser().parseFromString(xml, "text/xml");
+  const signatureNode = xpath.select1("//*[local-name(.)='Signature']", doc);
+  isDomNode.assertIsNodeLike(signatureNode);
+  const sig = new SignedXml({
+    publicCert: fs.readFileSync("./test/static/client_public.pem"),
+  });
+  sig.loadSignature(signatureNode);
+  return sig;
+}
+
+function resignOverWrappedSignedInfo(xml: string): string {
+  const doc = new xmldom.DOMParser().parseFromString(xml, "text/xml");
+  const signatureNode = xpath.select1("//*[local-name(.)='Signature']", doc);
+  isDomNode.assertIsNodeLike(signatureNode);
+  const sig = new SignedXml();
+  sig.loadSignature(signatureNode);
+  if (typeof sig.canonicalizationAlgorithm !== "string") {
+    throw new Error("canonicalizationAlgorithm missing after loadSignature");
+  }
+  if (sig.signatureAlgorithm == null) {
+    throw new Error("signatureAlgorithm missing after loadSignature");
+  }
+  const AlgoCtor = sig.SignatureAlgorithms[sig.signatureAlgorithm];
+  if (AlgoCtor == null) {
+    throw new Error(`signature algorithm '${sig.signatureAlgorithm}' is not supported`);
+  }
+  const signedInfo = xpath.select1(".//*[local-name(.)='SignedInfo']", signatureNode);
+  isDomNode.assertIsNodeLike(signedInfo);
+  const canon = sig.getCanonXml([sig.canonicalizationAlgorithm], signedInfo, {
+    ancestorNamespaces: findAncestorNs(doc, "//*[local-name(.)='SignedInfo']"),
+  });
+  const signatureValue = new AlgoCtor().getSignature(
+    canon,
+    fs.readFileSync("./test/static/client.pem"),
+  );
+  const resigned = xml.replace(
+    /(<SignatureValue>)([^<]*)(<\/SignatureValue>)/,
+    `$1${signatureValue}$3`,
+  );
+  if (resigned === xml) {
+    throw new Error("failed to replace SignatureValue");
+  }
+  return resigned;
+}
+
+function expectWrappedSignatureVerifies(element: string, uri: string, wrapped: string) {
+  const xml = resignOverWrappedSignedInfo(wrapAlgorithm(signBook(), element, uri, wrapped));
+  expect(xml.includes(`Algorithm="${wrapped}"`)).to.equal(true);
+  expect(load(xml).checkSignature(xml)).to.equal(true);
+}
+
+describe("xsd:anyURI whitespace collapse on Algorithm attributes", function () {
+  it("resolves CanonicalizationMethod when Algorithm is line-wrapped", function () {
+    const xml = wrapAlgorithm(signBook(), "CanonicalizationMethod", C14N, `${C14N}\n          `);
+    const sig = load(xml);
+
+    expect(sig.canonicalizationAlgorithm).to.equal(C14N);
+    expect(sig.signatureAlgorithm).to.equal(RSA_SHA1);
+    const ref = sig.getReferences()[0];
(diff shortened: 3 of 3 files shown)


cjbarth:
Thanks, Wes. This implements the normalization requested in #548, but review uncovered a concrete algorithm-policy bypass. We should hold this PR in its current form.

Reproduced against master `2ac129bc319fb78c578dfed42f6babb78b09e00d` and this PR at `dad853008a748a144cd828fb9a6b7da7be93b6c9`:

Start with a valid RSA-SHA1 signature and a consumer that rejects RSA-SHA1 by comparing the raw `SignatureMethod/@Algorithm` string with the registered URI. The consumer uses the same descendant lookup as `loadSignature`. Adding an unsigned `Object` before `SignedInfo`, containing a `SignatureMethod` with spaces around that URI, makes both lookups select the decoy. The original `SignedInfo` and signature value remain unchanged.

| Version | Consumer's raw-string check | Library verification |
| --- | --- | --- |
| Master | Allows the padded identifier | Throws `is not supported` |
| This PR | Allows the padded identifier | Collapses to RSA-SHA1; `checkSignature` returns `true` |
| Diagnostic-only alternative | Allows the padded identifier | Throws an error identifying the whitespace |

This reduces security for that consumer: the upgrade turns an attacker-modified document that was rejected into one that passes the combined policy and verification checks. The attacker needs an existing valid RSA-SHA1 signature under the verifier's trusted key, but does not need the signing key to insert the decoy. The demonstrated impact is bypassing the consumer's SHA-1 prohibition; the cryptographic signature still verifies over the original signed content.

The broad lookup predates this PR. Both algorithm selectors search descendants of `Signature` by local name, so unsigned content can supply the selected method. An exact, unpadded decoy is already honored on master, although the demonstrated consumer rejects it. The existing unsupported-identifier path fails closed; the description's earlier “failed open” wording was incorrect.

There is a further constraint on fixing this separately: an isolated experiment that only scopes the library's selectors to namespace-qualified direct children of `SignedInfo` still lets the same consumer be fooled. The consumer follows the decoy while the library verifies using the authentic method. Even a decoy naming SHA-256 then passes that consumer's check. A correction must address malformed structure and ambiguity as well as selector scope. This reproduction places `Object` before `SignedInfo`, contrary to the [specified Signature child order](https://www.w3.org/TR/xmldsig-core1/#sec-Signature).

Controls confirmed that checking the actual method directly under `SignedInfo` rejects this example, and removing RSA-SHA1 from the verifier's algorithm registry also prevents verification.

For #548, the immediate change should preserve rejection and explain the whitespace in the error. The existing selector/structure problem needs a separate security fix with public-API regression tests. Any later normalization proposal needs review against consumer algorithm policies as well as XML schema semantics.

markstos:
Further AI analysis:

A safe design needs three pieces:

1. **Reject ambiguous signature structure.** `loadSignature()` currently finds algorithm elements with descendant XPath queries, so unsigned content can shadow the real `SignedInfo` algorithms ([signed-xml.ts](/home/mark/git/xml-crypto/src/signed-xml.ts:624)). Parse the required direct children, validate cardinality/order, and reject extra or out-of-place `CanonicalizationMethod`/`SignatureMethod` elements. Merely ignoring the decoy is insufficient because a consumer may still select it.

2. **Make whitespace collapsing opt-in.** Preserve today’s fail-closed behavior by default, but offer an explicit option such as `algorithmAttributeWhitespace: "collapse"`. Canonicalization must still operate on the original DOM—the normalized value is only for algorithm resolution—so signed whitespace remains covered by `SignatureValue`.

3. **Require policy to use the normalized identifiers.** When collapsing is enabled, callers should supply an in-library algorithm policy/allowlist covering signature, digest, canonicalization, and transforms. The verifier would apply that policy to the same normalized values it later uses for registry lookup. For example, a policy excluding RSA-SHA1 must see normalized RSA-SHA1 and reject it before verification can succeed.

The existing registries around [signed-xml.ts](/home/mark/git/xml-crypto/src/signed-xml.ts:123) are already close to algorithm allowlists, but a constructor-level verification policy would make the security decision explicit and cover all four algorithm types consistently.

That produces these behaviors:

- Existing callers, without the option: padded algorithms remain rejected, ideally with whitespace made visible in the error.
- Opted-in callers with a suitable policy: genuinely signed, line-wrapped `Algorithm` values verify.
- The PR’s unsigned decoy: rejected as ambiguous/malformed structure.
- A normalized RSA-SHA1 value: rejected when the caller’s normalized policy forbids it.
- Whitespace inserted after signing: still fails cryptographic verification.

I would therefore split this into:

- A patch now: improved diagnostic plus structural/ambiguity regression tests.
- A separate feature, possibly v7: explicit normalization paired with verifier-owned algorithm policy and migration documentation.

There is no general way for the library to make unconditional normalization safe for arbitrary consumers that inspect the raw DOM themselves. The policy decision has to move inside the verifier—or normalization must remain an explicit compatibility opt-in. This directly addresses the concern documented in the [security review](https://github.com/node-saml/xml-crypto/pull/576#issuecomment-5639838236).

PR #591, opened 2026-09-14
fix: report a validated node or reference only after the signature verifies

cjbarth:
## Summary

Two public APIs reported unverified references as validated.

- `ref.getValidatedNode()` returned `null` only when that reference's digest failed. If the `SignatureValue` was forged, `checkSignature()` threw and `getSignedReferences()` stayed empty, but `getValidatedNode()` still returned the referenced node.
- `validateElementAgainstReferences()` compared the element's digest with the loaded, unverified `SignedInfo`. It returned the reference for a forged signature before `checkSignature()` ran, and again after it threw.

Both now consider a reference only once `checkSignature()` has verified the `SignatureValue`, which is when `ref.signedReference` is set. The `getValidatedNode()` deprecation warning now points only at `getSignedReferences()`, instead of also at `ref.signedReference`, which is only reachable through deprecated or unsafe APIs.

`getValidatedNode()` is deprecated, and `validateElementAgainstReferences()` is being deprecated in a separate PR. Following the rule that tests prove what the supported API does, this PR adds no test for either. It was verified with a script against a build:

| Case | `checkSignature()` | `getValidatedNode()` before / after | `validateElementAgainstReferences()` before / after |
| --- | --- | --- | --- |
| genuine signature | `true` | node / node | reference / reference |
| forged `SignatureValue` | throws | **node** / `null` | **reference** / throws "No references passed validation" |
| tampered content | `false` | `null` / `null` | not run |

The same `getValidatedNode()` result was confirmed on 6.1.2.

## Compatibility

Callers that ignore a failed `checkSignature()` and still use `getValidatedNode()` or `validateElementAgainstReferences()` get `null` or an error where they previously got content. Callers that check the result see no change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved signed XML validation for references that do not contain signed data.
  - Deprecated reference validation access now consistently returns no result when signed data is unavailable, preventing incorrect validation outcomes.
  - Updated guidance directs integrations to the supported signed-reference access method for retrieving validated signed references.


cjbarth:
Dropping this PR. Both methods it fixes are deprecated in 6.2: `ref.getValidatedNode()` already is, and #592 deprecates `validateElementAgainstReferences()`. Deprecated code is unsupported, so it gets no bug fixes and no tests. Use `getSignedReferences()` after `checkSignature()` instead.

